PR #24, opened 2021-01-14
[13.0][IMP] hr_expense_advance_clearing : check clearing on expense

Saran440:
- Add clearing on Expense.
- Check clearing on expense that you would clearing advance, it will required field `Clear Advance` on Expense Sheet.
- It will check clearing (if not advance and not clearing) on state Sumbit to Manager.

Selection_004
Selection_005

cc: @kittiu

### DIFF
--- a/hr_expense_advance_clearing/models/hr_expense.py
+++ b/hr_expense_advance_clearing/models/hr_expense.py
@@ -9,6 +9,7 @@ class HrExpense(models.Model):
     _inherit = "hr.expense"
 
     advance = fields.Boolean(string="Employee Advance", default=False)
+    clearing = fields.Boolean(default=False)
 
     @api.constrains("advance")
     def _check_advance(self):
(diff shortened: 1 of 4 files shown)


kittiu:
@Saran440 please add migration script for this version too. Because this improvement change database.
You need to ensure that, when user upgrade this module to the new version, the flag "clearing" is setup properly for existing documents.

kittiu:
If user select ( 2 clearing and 1 expense ) -> to create Report Sheet. The system will allow, but will set the 1 expenses to become clearing also.
I think it should either,
1. Not allow and give the error message "Selecting both expense and clearing is not allowed" (this has to be on both when create Report from Expenses, and on Report adding expense line).
2. Allow, but give warning message before proceeding "You are selecting expense with clearing. All expense will become clearing, do you want to proceed?".

WDYT?

Saran440:
Closed. you can see advance_sheet_id can specify clearing

PR #27, opened 2021-01-29
[14.0] [NEW] hr_expense_analytic_distribution: New module to distribute the expenses into various analytic account.

Chandresh-OSI:


Saran440:
Functional Test. I have question
1. If we don't access right Analytic Account. Should we can do this module?
2. we would be distribute to analytic 20% and and expense normal process 80%. Is it possible?
if not, you should constain total distribute must be 100

pedrobaeza:
Why not using analytic distributions from tags?

dreispt:
@Saran440 
1. Right now, no, you need at least to be able to select Analytic Account. But it would be simple to add that permission here.
2. It would be possible, but IMO error prone. I would prefer to require users to proves a 100% total distribution.

@pedrobaeza 
That was tested but doesn't work for the EE Finance Reports, such as P&L.
They basically ignore analytic tag distribution.

hr_expense_analytic_distribution/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,24 @@
...
+    count_percent = fields.Float(compute="_compute_count_percent")
+
+    @api.depends("expense_distribution_ids.percentage")
+    def _compute_count_percent(self):
+        percent = 0.0
+        for rec in self:
+            for line in rec.expense_distribution_ids:
+                percent += line.percentage
+                if percent > 100:
+                    raise UserError(_("Sorry, Percentage should be equal to 100."))
```
dreispt: Don't do validations in a computation.
Use a contrains method instead.

hr_expense_analytic_distribution/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,24 @@
...
+from odoo.exceptions import UserError
+
+
+class HrExpenseSheetInherit(models.Model):
+    _inherit = "hr.expense.sheet"
+
+    expense_distribution_ids = fields.One2many(
+        "hr.expense.analytic.distribution", "expense_id", string="Distribution Lines"
+    )
+    count_percent = fields.Float(compute="_compute_count_percent")
```
dreispt: Names are important. This is not a count, it is a sum or a total.
Actually, I don't think we even need this field.

hr_expense_analytic_distribution/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @api.depends("expense_distribution_ids.percentage")
+    def _compute_count_percent(self):
+        percent = 0.0
+        for rec in self:
+            for line in rec.expense_distribution_ids:
+                percent += line.percentage
+                if percent > 100:
+                    raise UserError(_("Sorry, Percentage should be equal to 100."))
+            rec.count_percent = 100 - percent
```
dreispt:
```suggestion
    @api.constrains("expense_distribution_ids.percentage")
    def _constrains_distribution_ids_percentage(self):
        for report in self:
            total_percent = sum(report.expense_distribution_ids.mapped("percentage")
            if total_percent != 100:
                    raise UserError(_("Sorry, Percentage should be equal to 100."))
```

hr_expense_analytic_distribution/models/hr_expense.py
```diff
@@ -0,0 +1,130 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrExpense(models.Model):
+
+    _inherit = "hr.expense"
+
+    def _get_account_move_line_values(self):
```
dreispt, line 10: The right way to do this is to call super() and get the result.
Then work on that data structure, copying and modifying the line data to perform the distribution.
Not using super() breaks the inheritance chain.

dreispt: I would love to see this implemented in under 40-50 LoC

dreispt (changes requested):

pedrobaeza:
@dreispt financial reports has nothing to do with this. I'm talking about adding to the expense tags with distribution, and make them to generate analytic entries.

dreispt:
@pedrobaeza Yes, ff you don't care about the financial reports, that works and you don't need this feature.
In my case, I expect to be able to analyze the P&L by Analytic Account, and Analytic Tags don't work for that.

pedrobaeza:
Yes, but I'm not talking about that. I'm talking about not reinventing a new model for reflecting distributions: use tags model, but do the same split over accounting entries.

dreispt:
@pedrobaeza I see your point.

We expect distribution to be different in most expense reports.
So the UX would be either:
a) have the user search Tags for an existing one, and create one if if doesn't exist. This is hard for basic expense users.
b) create a tag for each expense report. This would pollute the tags list, in case they are being used for other purposes. 

The above the UX inconveniences made me prefer a manual distribution for each expense report.
What do you think?

pedrobaeza:
OK, I got as well your point. Maybe you can do the following: on the distribution part, allow to select an analytic tag with distribution, and on such onchange, populate your distribution lines with such information. So you get the best of both worlds, having the option to use "templates" for distributions, or create them manually.

dreispt:
@pedrobaeza That sounds nice, but It is really two feature working together, and I don't want to implement the two.
I could extend the Tag to also impact Accounting Moves, and always use a Tag on the Expense Report, that could easily be created on the fly. That solves the Expense Report UX.

I worry that:
a) the additional Account Move splits on Tags would have collateral effects on other usages of it, and 
b) the big number of Tags this process would create.

There are ways to mitigate b), like creating archived entries, or use a flag filter.
Is a) something to worry about?

pedrobaeza:
I think you haven't understood me: I'm only proposing to have a field tags to act as template for your distribution, isolated from any other tag use, but anyway, don't worry if you don't want to implement this. It was just a suggestion, but not real case for me for now.

dreispt:
I see - the actual split logic would still be independent from Tags.
That is a good suggestion, I'll keep it in mind, it could be an enhancement if the template use case is brought to the table.
Thank you Pedro.

dreispt (approved):
I code reviewed, and in the process added tests and streamlined the business logic.

PR #28, opened 2021-02-03
[13.0][FIX] hr_expense_invoice: Set groups to invoice_id field in expense view

victoralmau:
Set groups to `invoice_id` field in expense view

@Tecnativa TT28121


pedrobaeza (approved):

pedrobaeza:
Same as in v12, so fast-tracking:

/ocabot merge minor

PR #31, opened 2021-03-23
[13.0][FIX+IMP] hr_expense_invoice: Don't mark the sheet as paid + improve reconciliation

victoralmau:
- Don't mark the sheet as paid (paid by employee)
- Do reconciliations per line, so don't group invoices from different partners.
- Intercept properly the sheet paid action for not marking it as paid when
  reconciling the expense lines that belongs to invoices.
- Some docstring.
- Hide taxes in expense when invoice is defined (to prevent errors).
- Prevent set invoices paid when sheet paid by company.

Please @pedrobaeza, can you review it ?

@Tecnativa TT28869


pedrobaeza (approved):
Something about "Paid by company"?

victoralmau:
By default in Odoo when an expense is paid by the company, when publishing the expense it is marked as paid and a payment is created (`account.payment`). With this addon the operation is the same, do you think you have to change it?

pedrobaeza:
@victoralmau the question is if the invoices are marked as paid, which is an error. I also doubt if the payment should contain the amount of the invoices or not.

victoralmau:
Yes, invoices are marked as paid. Do you think that it's not correct and need to appear as unpaid? (not auto-reconcile)
Amount payment is defined acording to odoo core here: https://github.com/odoo/odoo/blob/13.0/addons/hr_expense/models/hr_expense.py#L497

pedrobaeza:
No, they shouldn't be marked as paid, as they are still pending. I would simply forbid for now to post expense sheet in this case and add it to the roadmap.

victoralmau:
Ok, perfect.
Added some changes to prevent set invoices as paid when sheets is paid by company  and add test about it.

kittiu:
Function tested, LGTM.

I also found that, now it is allowed to choose the Paid invoice too.

image

PR #32, opened 2021-03-26
[13.0][FIX+imp] hr_expense_invoice: Set empty tax_ids when create_invoice wizard + Remove expense_create_invoice wizard according to sheet

victoralmau:
[FIX] Set empty `tax_ids` when create_invoice wizard.
[IMP] Remove expense_create_invoice wizard according to sheet

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT28869


pedrobaeza:
Why didn't you cherry-pick my v12 commit?

victoralmau:
It's different about https://github.com/OCA/hr/commit/3bf221e25e735be2c9cb341236a8310cdad754bf (although it's need to apply 12.0 too)

pedrobaeza:
But the other exists in v13?

victoralmau:
Yes, in 13.0 exists in https://github.com/OCA/hr-expense/commit/95f1cff925afc61bc4aea302ba474248660f39c2

pedrobaeza:
OK, are you going to add the constraint?

victoralmau:
> OK, are you going to add the constraint?

IMO is not necessary.

hr_expense_invoice/wizard/expense_create_invoice.py
```diff
@@ -62,5 +62,5 @@ def create_invoice(self):
         invoice = (
             self.env["account.move"].with_context(type="purchase").create(invoice_vals)
         )
-        self.expense_ids.write({"invoice_id": invoice.id})
+        self.expense_ids.write({"invoice_id": invoice.id, "tax_ids": False})
```
pedrobaeza: What about the final amount? Maybe it has been changed.

victoralmau: Is not possible because total of sheet is computed according to lines (expenses).

pedrobaeza: Not total of sheet, but the line.

I see this very weird... Are you saying that the invoice is assigned to all the lines? That's a mistake IMO.

victoralmau: This is done through a wizard when the lines have been selected, so it is possible to assign the same invoice generated to several lines and then, the total cannot be assigned correctly.

pedrobaeza: At least a constraint should be added to raise the error if the total amount doesn't match. If not, there will be problems. @kittiu what is your use case for adding such weird wizard?

kittiu: With my blurry memory. In v8, we are used to a module that create invoice from expense.
But to be frank, nowadays we didn't use it either. Please feel free to remove. It is erroneous to have too many ways to do the same.

pedrobaeza (approved):

PR #34, opened 2021-04-07
[13.0][IMP+FIX] hr_expense_invoice: Add button to create invoice from expenses + Allow create invoice from invoice_id field

victoralmau:
[IMP] Add button to create invoice from expenses.
[FIX] Allow create invoice from invoice_id field

In these PR https://github.com/OCA/hr-expense/pull/32 we remove it option to create invoice from some expenses but it's a interesting functionallity and with these change add button in expense to allow create invoice with only 1 line.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT29072

### DIFF
--- a/hr_expense_invoice/models/account_move.py
+++ b/hr_expense_invoice/models/account_move.py
@@ -1,10 +1,31 @@
 # Copyright 2019 Ecosoft <saranl@ecosoft.co.th>
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+from odoo.tools import float_compare
 
 
 class AccountMove(models.Model):
     _inherit = "account.move"
+
+    expense_ids = fields.One2many(
+        comodel_name="hr.expense", inverse_name="invoice_id", string="Expenses"
+    )
+
+    @api.constrains("amount_total")
+    def _check_expense_ids(self):
+        for move in self.filtered("expense_ids"):
+            DecimalPrecision = self.env["decimal.precision"]
+            precision = DecimalPrecision.precision_get("Product Price")
+            expense_amount = sum(move.expense_ids.mapped("total_amount"))
+            if float_compare(expense_amount, move.amount_total, precision) != 0:
+                raise ValidationError(
+                    _(
+                        "You can't change the total amount, as there's an expense "
+                        "linked to this invoice."
+                    )
+                )
 
     def _get_cash_basis_matched_percentage(self):
         res = super()._get_cash_basis_matched_percentage()
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -19,6 +19,7 @@ class HrExpense(models.Model):
             ("type", "=", "in_invoice"),
             ("state", "=", "posted"),
             ("invoice_payment_state", "=", "not_paid"),
+            ("expense_ids", "=", False),
         ],
         copy=False,
     )
@@ -95,5 +96,7 @@ class HrExpense(models.Model):
         """
         if self.invoice_id:
             self.quantity = 1
-            self.unit_amount = self.invoice_id.amount_total
             self.tax_ids = [(5,)]
+            # Assign this amount after removing taxes for avoiding to raise
+            # the constraint _check_expense_ids
+            self.unit_amount = self.invoice_id.amount_total
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -11,7 +11,7 @@ from odoo.exceptions import UserError
 class HrExpense(models.Model):
     _inherit = "hr.expense"
 
-    sheet_id_state = fields.Selection(related="sheet_id.state")
+    sheet_id_state = fields.Selection(related="sheet_id.state", string="Sheet state")
     invoice_id = fields.Many2one(
         comodel_name="account.move",
         string="Vendor Bill",
@@ -44,7 +44,6 @@ class HrExpense(models.Model):
             .with_context(default_type="in_invoice")
             .create(
                 {
-                    "type": "in_invoice",
                     "ref": self.reference,
                     "invoice_date": self.date,
                     "invoice_line_ids": invoice_lines,
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -1,6 +1,7 @@
 # Copyright 2015-2020 Tecnativa - Pedro M. Baeza
 # Copyright 2017 Tecnativa - Vicent Cubells
 # Copyright 2020 Tecnativa - David Vidal
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
@@ -10,6 +11,7 @@ from odoo.exceptions import UserError
 class HrExpense(models.Model):
     _inherit = "hr.expense"
 
+    sheet_id_state = fields.Selection(related="sheet_id.state")
     invoice_id = fields.Many2one(
         comodel_name="account.move",
         string="Vendor Bill",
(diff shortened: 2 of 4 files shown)


victoralmau:
Any idea to create 1 invoice according to multi-expense with "one click"?
Maybe a use case about it is a sheet example with 10 expenses and we have to create only 1 invoice with 10 lines (line per expense).

hr_expense_invoice/models/hr_expense.py
```diff
@@ -21,6 +23,37 @@ class HrExpense(models.Model):
...
+                    "tax_ids": [(6, 0, self.tax_ids.ids)],
+                },
+            )
+        ]
+        invoice = (
+            self.env["account.move"]
+            .with_context(default_type="in_invoice")
+            .create(
+                {
+                    "type": "in_invoice",
```
joao-p-marques: Why do you need to specify the type twice? I'd say the correct one would be through the context.

hr_expense_invoice/tests/test_hr_expense_invoice.py
```diff
@@ -232,6 +232,39 @@ def test_2_hr_test_multi_invoices(self):
...
+        self.expense2.action_expense_create_invoice()
+        self.assertTrue(self.expense2.invoice_id)
+        self.assertEqual(self.sheet.invoice_count, 2)
+        # Validate invoices
+        self.expense.invoice_id.partner_id = self.partner
+        self.expense.invoice_id.action_post()
+        self.expense2.invoice_id.partner_id = self.partner
+        self.expense2.invoice_id.action_post()
+        self.sheet.with_context(
+            {"default_expense_line_ids": self.expense.id}
```
joao-p-marques: Are you sure this is correct? Isn't `default_expense_line_ids` expecting a list of ids?

joao-p-marques (commented):
Code review

joao-p-marques (approved):

hr_expense_invoice/views/hr_expense_views.xml
```diff
@@ -41,6 +44,18 @@
...
+            <xpath
+                expr="//field[@name='expense_line_ids']/tree/field[@name='total_amount']"
+                position="after"
+            >
+                <field name="sheet_id_state" invisible="1" />
+                <button
+                    name="action_expense_create_invoice"
+                    string="Create Vendor Bill"
+                    type="object"
+                    attrs="{'invisible': ['|', ('invoice_id', '!=', False),('sheet_id_state', '!=', 'approve')]}"
```
pedrobaeza: Shouldn't you allow it on draft as well?

hr_expense_invoice/models/account_move.py
```diff
@@ -1,11 +1,26 @@
...
     _inherit = "account.move"
 
+    expense_ids = fields.One2many(
+        comodel_name="hr.expense", inverse_name="invoice_id", string="Expenses"
+    )
+
+    @api.constrains("amount_total")
+    def _check_expense_ids(self):
+        for move in self:
+            if move.expense_ids:
```
pedrobaeza: Save one indentation level:
```suggestion
        for move in self.filtered("expense_ids"):
```

hr_expense_invoice/models/account_move.py
```diff
@@ -1,11 +1,26 @@
...
+                if sum(move.expense_ids.mapped("total_amount")) != move.amount_total:
+                    raise UserError(
```
pedrobaeza: On constraints, it should be `ValidationError`:
```suggestion
                    raise ValidationError(
```

hr_expense_invoice/models/account_move.py
```diff
@@ -1,11 +1,26 @@
...
+                        _("Total amount can't change because some expense is related")
```
pedrobaeza:
```suggestion
                        _("You can't change the total amount, as there's an expense linked to this invoice.")
```

hr_expense_invoice/models/account_move.py
```diff
@@ -1,11 +1,26 @@
...
 
+    expense_ids = fields.One2many(
+        comodel_name="hr.expense", inverse_name="invoice_id", string="Expenses"
+    )
+
+    @api.constrains("amount_total")
+    def _check_expense_ids(self):
+        for move in self:
+            if move.expense_ids:
+                if sum(move.expense_ids.mapped("total_amount")) != move.amount_total:
```
pedrobaeza: Use `float_compare` for avoiding rounding problems.

hr_expense_invoice/models/hr_expense.py
```diff
@@ -56,5 +89,5 @@ def _onchange_invoice_id(self):
         """
         if self.invoice_id:
             self.quantity = 1
-            self.unit_amount = self.invoice_id.amount_total
             self.tax_ids = [(5,)]
+            self.unit_amount = self.invoice_id.amount_total
```
pedrobaeza, line 102: Why this change or order?

pedrobaeza (changes requested):

hr_expense_invoice/models/hr_expense.py
victoralmau: It's very important because if assign `unit_amount` before se `tax_ids` to empty, invoice constraint show (`_check_expense_ids`). 
These little change it's necessary to "_Create Vendor Bill_" button in expense work fine.

pedrobaeza: I'm not sure to get you. Are you saying that a constraint raises? Which one? In any case, if so, please put a comment like:
```python
# Assign this amount after removing taxes for avoiding to raise the constraint X
```
and remember always that when something is not obvious, you should put a comment for that.

hr_expense_invoice/models/hr_expense.py
```diff
@@ -10,17 +11,49 @@
...
+            .with_context(default_type="in_invoice")
+            .create(
+                {
+                    "ref": self.reference,
+                    "invoice_date": self.date,
+                    "invoice_line_ids": invoice_lines,
+                }
+            )
+        )
+        self.write({"invoice_id": invoice.id, "tax_ids": False})
```
pedrobaeza: I miss here quantity = 1 and unit_amount = total invoice

pedrobaeza (changes requested):

hr_expense_invoice/models/hr_expense.py
```diff
@@ -10,17 +11,56 @@
...
+    def action_expense_create_invoice(self):
+        invoice_lines = [
+            (
+                0,
+                0,
+                {
+                    "product_id": self.product_id.id,
+                    "name": self.name,
+                    "price_unit": self.unit_amount,
+                    "quantity": 1,
```
pedrobaeza: This change is not correct AFAIK

hr_expense_invoice/views/hr_expense_views.xml
```diff
@@ -5,18 +5,24 @@
...
-                            'default_reference': reference,
-                            'default_date_invoice': date,
-                            'default_invoice_line_ids': [{'product_id': product_id,
+                            'default_ref': reference,
+                            'default_invoice_date': date,
+                            'default_line_ids': [{'product_id': product_id,
                                                           'name': name,
                                                           'price_unit': unit_amount,
-                                                          'quantity': quantity,
+                                                          'quantity': 1,
```
pedrobaeza: Neither this one

hr_expense_invoice/models/hr_expense.py
victoralmau: These changes it's about data consistency .
If we change `quantity=1` in `invoice_id` onchange, IMO it's correct to set exactly the same in other use cases.
What do you think about it?

pedrobaeza: No, it isn't. Starting with, you are not multiplying the unit_amount * quantity in the price_unit, but the question is the consistency is needed from invoice to expense, not the reverse. Please leave it as it was.

victoralmau: Sorry but if leave it as it was these comment https://github.com/OCA/hr-expense/pull/34#discussion_r615237666 will be pending right?

pedrobaeza: This one still remains incorrect

pedrobaeza (approved):

PR #41, opened 2021-05-16
[14.0][MIG] hr_expense_advance_clearing_sequence

Saran440:
Change sequence EXAV -> AV
- [x] test script


kittiu:
@Saran440 we also have this module to make sequence more flexible. https://github.com/OCA/server-tools/pull/2080

Saran440:
@kittiu I think this module add 1 function. it's not necessary to inherit it.
However, if hr_expense_advance_clearing has clearing field on hr.expense,
I will enhanced this module depend on it.
WDYT?

kittiu:
> @kittiu I think this module add 1 function. it's not necessary to inherit it.
> However, if hr_expense_advance_clearing has clearing field on hr.expense,
> I will enhanced this module depend on it.
> WDYT?

This module can MIG as normal.

hr_expense_advance_clearing_sequence/hooks.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2021 Ecosoft Co., Ltd (http://ecosoft.co.th/)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, api
+
+
+def assign_old_sequences(cr, registry):
```
kittiu: Why are we overwriting the new number to already assigned number?
I think we shouldn't change the existing number, should we?

Saran440: because advance document should number AVEX instead of EX, or not?
but EX sequence will skip this number. I'm not sure what is best way, Could you suggest?

kittiu: First small thing. I think the number should be AVxxxx (not AVEX)
I am not sure when you say, EX will skip the number? But I am sure, we shouldn't overwrite the existing number.

Saran440: Okay, I see.

hr_expense_advance_clearing_sequence/data/hr_expense_data.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <record id="seq_expense_advance" model="ir.sequence">
+        <field name="name">Expense report sequence (Advance)</field>
+        <field name="code">hr.expense.sheet.advance</field>
+        <field name="padding" eval="4" />
+        <field name="prefix">EXAV</field>
```
kittiu:
```suggestion
        <field name="prefix">AV</field>
```
I think AV is enough.

kittiu:
@Saran440 this module can be simplified as default_advance is now used in `hr_expense_advance_clearing`
https://github.com/OCA/hr-expense/pull/41/commits/9418ae3f703de717bc6b154f378230e23c29d883

If you are ok, I think you should squash commit back to 1 commit. Thanks!

kittiu:
If ok, you may need to fix test script too, as this one is more simplified.

hitrosol (approved):
Functional test okay...

ps-tubtim (approved):

PR #46, opened 2021-07-12
[14.0][ADD] hr_expense_exception

kittiu:
TODO:
- [x] Test


Saran440 (approved):
Functional Test :+1: 

Should we add function `action_sheet_move_create` (Post Journal Entries) on exception?
because Some case we need check exception before post accounting.

newtratip (approved):
LGTM :+1:

Issue #44, opened 2021-06-13
[14.0] hr_expense_advance_clearing, need code review for "return advance" process.


kittiu:
There is a part of code that is nor correctly written.

https://github.com/OCA/hr-expense/blob/14.0/hr_expense_advance_clearing/wizard/account_payment_register.py#L34-L39

It is trying to avoid a warning on default_get() by update account type back and forth.

kittiu:
cc @Saran440 @newtratip

kittiu:
@Saran440 can you fix this please?

Saran440:
Fixed.

PR #47, opened 2021-07-19
[FIX] hr_expense_advance_clearing: default_get return advance

Saran440:
fix from issue https://github.com/OCA/hr-expense/issues/44

In v14 odoo merged register payment wizard on expense into account.payment.register and check some condition
So, expense return advance can't use register payment wizard.

I check context and override default_get on wizard, if it from return advance only.

cc: @kittiu

### DIFF
--- a/hr_expense_advance_clearing/wizard/account_payment_register.py
+++ b/hr_expense_advance_clearing/wizard/account_payment_register.py
@@ -4,40 +4,72 @@
 from werkzeug.urls import url_encode
 
 from odoo import _, api, models
+from odoo.exceptions import UserError
 
 
 class AccountPaymentRegister(models.TransientModel):
     _inherit = "account.payment.register"
 
-    @api.model
-    def default_get(self, fields_list):
-        """Step to avoid UserError from core odoo:
-        - Change account type to receivable
-        - Call super() and rollback account to origin
-        """
-        advance_return = self._context.get("hr_return_advance", False)
-        origin_account = False
-        if advance_return and self._context.get("active_model") == "account.move":
+    def _get_default_advance(self):
+        """ This function is default_get from return advance only """
+        return {}
+
+    def _default_return_advance(self, fields_list):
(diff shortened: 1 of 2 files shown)


hr_expense_advance_clearing/wizard/account_payment_register.py
```diff
@@ -4,41 +4,73 @@
...
-        """Step to avoid UserError from core odoo:
-        - Change account type to receivable
-        - Call super() and rollback account to origin
-        """
-        advance_return = self._context.get("hr_return_advance", False)
-        origin_account = False
-        if advance_return and self._context.get("active_model") == "account.move":
+    def _get_default_advance(self):
+        """ This function is default_get from return advance only """
+        return {}
```
kittiu, line 15:
```suggestion
```

hr_expense_advance_clearing/wizard/account_payment_register.py
```diff
@@ -4,41 +4,73 @@
...
+
+    def _default_return_advance(self, fields_list):
+        """ OVERRIDE: lines without check account_internal_type for return advance only """
+        res = self._get_default_advance()
```
kittiu, line 19:
```suggestion
        res = {}
```

hr_expense_advance_clearing/wizard/account_payment_register.py
kittiu: I think this is not need. Please discuss if you see why we need this.

kittiu (approved):

kittiu:
Thanks for this fix. It think this is the most we can do already.

hr_expense_advance_clearing/wizard/account_payment_register.py
Saran440: this function for other module need add default some field, when you return advance.

kittiu: OK.

PR #49, opened 2021-07-20
[14.0][IMP] hr_expense_advance_clearing, add invisible fields

kittiu:
Add fields as invisible, this is not a harmful change, but can be useful when we pass default context of these field.
* advance
* product_id

@Saran440 can you FFW please.


Saran440 (approved):
LGTM :+1:

PR #53, opened 2021-07-23
[14.0][ADD] hr_expense_widget_o2m

kittiu:
Currently, if user create new expense sheet first, and then click to add expense lines,
System will popup a selection window (many2many widget) allowing user to select the existing expenses first, and also to create new expense.

This can be confusing for some user.

With this module, user have option to directly add new expense lines on expense report (one2many widget).

This module bring back the classic way to add new expense line.

image

Change from,

image

To,

image


kittiu:
cc @pedrobaeza @Saran440 

What do you think? Also the naming of module.

pedrobaeza:
Well, not sure it's really needed, as with current 14.0 way, you avoid 2 sources of expenses: the normal menu, and the direct addition.

kittiu:
> you avoid 2 sources of expenses

Thanks for the comment.

I do understand how the current 14.0 way works and normally use that for most customers. But in our case, we have the need (for some longer reasons), to access only Expense Sheet (direct addition). And with existing widget="many2many_tags", user sometime create the expense from that widget, but not used and don't know how to delete.

Saran440:
@kittiu I'm okay.
~~May be, Should this module add smart button link to expense?~~ Not necessary to do.

hr_expense_widget_o2m/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,20 @@
+<odoo>
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="name">res.config.settings.view.form.inherit.hr.expense</field>
+        <field name="model">res.config.settings</field>
+        <field name="priority" eval="85" />
+        <field name="inherit_id" ref="hr_expense.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[@name='expenses_setting_container']" position="inside">
+                <div class="col-12 col-lg-6 o_setting_box" id="uom_settings">
```
Saran440:
```suggestion
                <div class="col-12 col-lg-6 o_setting_box" id="expense_widget_o2m_settings">
```

Saran440 (commented):
Code Review:
name is duplicated

Saran440:
@kittiu I delete view on hr.expense Could you review this again, please?

kittiu:
@Saran440 this works?

Saran440:
Yes, it's work for me because this module can fixed some issue m2m widget (core odoo)

Saran440 (approved):
Functional Test :+1:

nimarosa (approved):

Issue #52, opened 2021-07-22
[RFC] expense sheet link to payment


Saran440:
Hello all,

In OCA has module `hr_expense_cancel` and it add new field `expense_sheet_id` in account payment.
I think this field should split out to new module **hr_expense_payment** because some module need to do something on payment and link to expense sheet.

Concept:
- Module `hr_expense_payment` will link expense sheet id at view payment.
- Module `hr_expense_cancel` depend on `hr_expense_payment`

What do you think?

PR #55, opened 2021-07-27
[14.0][ADD] hr_expense_payment (related to hr_expense_cancel)

Saran440:
From issue https://github.com/OCA/hr-expense/issues/52

module `hr_expense_cancel` add new field `expense_sheet_id` in account payment.
and it not support for case multi expense sheet register 1 payment.

This PR add new module `hr_expense_payment` and delete field expense on `hr_expense_cancel`

- `hr_expense_payment` add relation between expense sheet (report) and payment many2many.
it support case
    - Multi expense sheet - 1 payment
    - 1 expense sheet - Multi payment

- `hr_expense_cancel` move on field expense and hooks.py and depend on `hr_expense_payment`


kittiu:
@pedrobaeza @ernestotejeda are you ok with this change, as it affect `hr_expense_cancel` too.
But for other module to depends, I think it will be more useful.

pedrobaeza:
OK, but please include the migration script that transfers data from old many2one to the new x2m field.

Saran440:
@pedrobaeza done, thank you for reviewed.

kittiu:
@pedrobaeza I think this module is ready for sometime. Would you take a look please? Thanks!

ps-tubtim (approved):
Functional Test: Ok.

kittiu:
@Saran440 can you check travis please? May be some rebase will do.

kittiu (approved):

PR #59, opened 2021-08-01
[14.0][ADD] hr_expense_request_flow, hr_expense_advance_clearing_request_flow

kittiu:
Depends on:

- [ ] https://github.com/OCA/server-tools/pull/2150
- [ ] Test


hr_expense_advance_clearing_request_flow/models/hr_expense.py
```diff
@@ -0,0 +1,68 @@
...
+        # Request av must have default account depend on product advance
+        if self._context.get("use_av", False):
+            emp_advance = self.env.ref(
+                "hr_expense_advance_clearing.product_emp_advance"
+            )
+            return emp_advance.property_account_expense_id
+        return super()._default_account_id()
+
+    account_id = fields.Many2one(
+        comodel_name="account.account",
```
kittiu, line 58: Not sure we need this line?
```suggestion
```

Saran440: Not necessary, I need inherit default function only.

PR #62, opened 2021-08-12
[14.0][ADD] hr_expense_payment_term

Saran440:
This module allow expense sheet post journal entries with payment term similar invoice menu

How to use this module
1. you can configure accounting date on expense by current date.
when you create expense but post journal entries another day, accounting date should be current date.
Selection_001

2. Add new field `Due Date` on expense sheet similar invoice document
when you post journal entries, it will add due date into journal entries line.
Example: accounting date 12 Aug 2021, Due Date 15 days
Selection_003

Due Date on Journal Entries Line will 27 Aug 2021
Selection_004


Saran440:
https://github.com/OCA/hr-expense/pull/63

PR #64, opened 2021-08-18
[14.0][IMP] hr_expense_advance_cleaing: show advance on employee view

Saran440:
- Add advance tab on employee
Selection_001

- Add amount to clear on tree view expense report, it helpful user to find amount residual
Selection_002

### DIFF
--- a/hr_expense_advance_clearing/__manifest__.py
+++ b/hr_expense_advance_clearing/__manifest__.py
@@ -9,7 +9,12 @@
     "license": "AGPL-3",
     "website": "https://github.com/OCA/hr-expense",
     "depends": ["hr_expense"],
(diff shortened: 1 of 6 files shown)


pedrobaeza:
On the employee form, instead of a new tab that makes to load the form slower, you should use an smart-button as common technique for linking things.

Saran440:
okay, it's make sense. thanks :)

albariera:
Could the button be invisible when the advance count is 0?

Saran440:
@albariera Fixed.

albariera (approved):
Functional and technical review LGTM :+1:

kittiu (approved):

PR #65, opened 2021-08-20
[14.0][FIX] hr_expense_tier_validation - state submit

Saran440:
Core odoo change it back to submit when you create report from expense.
So, tier validation should tier submit.

https://github.com/odoo/odoo/commit/45d1b79fa779f8e890779417919e4fed85730cb2#diff-115a12ac273289d3b8f9698bbbbde87b72143b776c17b1ce9e5b465b0a8aff1c

### DIFF
--- a/hr_expense_tier_validation/models/hr_expense_sheet.py
+++ b/hr_expense_tier_validation/models/hr_expense_sheet.py
@@ -7,7 +7,7 @@ from odoo import models
 class HrExpenseSheet(models.Model):
     _name = "hr.expense.sheet"
     _inherit = ["hr.expense.sheet", "tier.validation"]
-    _state_from = ["draft", "submit"]
-    _state_to = ["approve", "post", "done"]
+    _state_from = ["draft"]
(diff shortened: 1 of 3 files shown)


albariera:
Functional review LGTM but two questions:
- Should the menu entry "Reports to Submit" be added at "Expense Reports" too?
- Could the button "Submit to Manager" be hidden if tier validation is needed before?
Thanks :smile:

Saran440:
@albariera Thanks for your reviewed.

Should the menu entry "Reports to Submit" be added at "Expense Reports" too?
- Report expense state `draft` for user check information. then, user can see report is draft from "My Reports"  BUT manager can see all report that them need to approve. So, menu "Reports to Approve" is first for manager want to see. 

Could the button "Submit to Manager" be hidden if tier validation is needed before?
- It shouldn't hide button because Case tier validation with myself it can skip tier validation to submit manager.

albariera:
I see your point! Thank you :smiley:

albariera (approved):

ps-tubtim (approved):

PR #78, opened 2021-11-09
[14.0][FIX] hr_expense_advance_clearing: constrains with advance clearing

Saran440:
Fix 2 case

- When the advance document's choose `account` does not equal the `expense account` (Product Advance), you cannot clear or return it.

>     **How to fix**
>     - Add constrains account on advance and Expense Account must be the same account

--------------------------------------------------------------

- Multi advance can't return advance.

Step to error
1. Create advance and add expense line multi (i.e. 10k and 10k)
2. Advance with normal process
3. Return Advance -> 15k (or more than 10k)
4. Return Advance again. it will error

>     **How to fix**
>     - Filter move line is not reconciled only

cc: @kittiu

### DIFF
--- a/hr_expense_advance_clearing/models/hr_expense.py
+++ b/hr_expense_advance_clearing/models/hr_expense.py
@@ -40,6 +40,10 @@ class HrExpense(models.Model):
                 raise ValidationError(
                     _("Employee advance, selected product is not valid")
                 )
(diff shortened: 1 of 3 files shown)


kittiu (approved):

ps-tubtim (approved):
It's worked for me.

PR #81, opened 2021-11-11
[14.0][ENH] hr_expense_advance_clearing, better expense_post_return_advance()

kittiu:
This ENH did nothing, just having function expense_post_return_advance() return something useful.

cc @Saran440 @ps-tubtim


Saran440 (approved):
No major change :+1:

PR #82, opened 2021-11-11
[14.0][ADD] hr_expense_sheet_payment_state

yankinmax:
Backport of the Payment Status (payment_state field) in Expense Report which is introduced in Odoo Community Version 15.0
https://github.com/odoo/odoo/commit/bb1f14e728e31de9f6783f2786b0af6f5db07c7e

Depends on:
- [ ] https://github.com/OCA/hr-expense/pull/146


hr_expense_sheet_payment_state/models/account_move.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def is_invoice(self, include_receipts=False):
+        # override needed to ensure payment_state is computed
```
grindtildeath:
```suggestion
        # Override move type check in the context of payment_state computation,
        # because the move linked to an expense sheet is of type entry.
```

hr_expense_sheet_payment_state/models/hr_expense.py
```diff
@@ -0,0 +1,44 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class HrExpense(models.Model):
+    _inherit = "hr.expense"
+
+    def _get_account_move_line_values(self):
+        result = super()._get_account_move_line_values()
```
grindtildeath, line 14:
```suggestion
        # As the move linked to an expense sheet is of type entry, we need
        # to manually exclude the payable line from 'invoice tab' to ensure
        # the payment_state is computed correctly as it is on vendor bills.
        result = super()._get_account_move_line_values()
```

grindtildeath (approved):
LGTM, I added two suggestion as comments to explicit a bit more why the overrides are needed.

hr_expense_sheet_payment_state/models/account_move.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def is_invoice(self, include_receipts=False):
+        # Override move type check in the context of payment_state computation,
+        # because the move linked to an expense sheet is of type entry.
+        self.ensure_one()
+        if self.line_ids.expense_id:
```
i-vyshnevska: line_ids o2m cannot be treated like that

hr_expense_sheet_payment_state/models/hr_expense.py
```diff
@@ -0,0 +1,47 @@
...
+        readonly=True,
+        copy=False,
+        tracking=True,
+        compute="_compute_payment_state",
+    )
+
+    @api.depends("account_move_id.payment_state")
+    def _compute_payment_state(self):
+        for sheet in self:
+            sheet.payment_state = sheet.account_move_id.payment_state or "not_paid"
```
i-vyshnevska, line 47: this is repeated functionality of related field

i-vyshnevska (changes requested):

yankinmax:
@i-vyshnevska take a look pls here:
https://github.com/odoo/odoo/commit/bb1f14e728e31de9f6783f2786b0af6f5db07c7e
I tried to implement the same. If you have suggestions pls welcome. This is not as easy as it can seem from the first POV (we discussed with @grindtildeath  that fact).

i-vyshnevska:
> @i-vyshnevska take a look pls here: [odoo/odoo@bb1f14e](https://github.com/odoo/odoo/commit/bb1f14e728e31de9f6783f2786b0af6f5db07c7e) I tried to implement the same. If you have suggestions pls welcome. This is not as easy as it can seem from the first POV (we discussed with @grindtildeath that fact).

implementations from odoo are not always the best practices I don't see an issue not having a related field as it will drop  ~10 lines of code, if there is an issue that it should be noted in the comment of compute method

hr_expense_sheet_payment_state/models/hr_expense.py
yankinmax: ok, I agree with you it's more explicit to add related, but I don't know how to combine related and that "or not_paid". Can you pls help?

grindtildeath: Actually it's a bit more than that as it provides a fallback value using `or` in case the move doesn't exist yet...

hr_expense_sheet_payment_state/models/account_move.py
grindtildeath: Actually looks like this is already possible in v14.0 (I also thought it was v.15 only)

grindtildeath (commented):
@yankinmax As this is a backport from Odoo code, you must still change license to LGPL to respect the original one.

hr_expense_sheet_payment_state/models/account_move.py
i-vyshnevska: ok I didn't know about that, even so on implementation for v15 this check is done in a separate hook and method is_invoice is not overridden completely, this is not very safe
anyhow build should be fixed

yankinmax: About the build.
We have an error on field which isn't installed to db by our module.
All the modules are installed during test instance is prepared.
I tried to run tests specifying different tags to tagged decorator to avoid testing on the fields which aren't related to my module, but still has an error. As it's not related to my module, we tested locally with @grindtildeath and left it as it's.

grindtildeath: @yankinmax I guess to make this safer, you could add a context key on the `_compute_amount` function since both calls to `is_invoice` were replaced by calls to `_payment_state_matters`, and check the existence of the context key before returning `True`.

yankinmax: @i-vyshnevska does it look safer now?

hr_expense_sheet_payment_state/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def is_invoice(self, include_receipts=False):
+        # Override move type check in the context of hr.expense.sheet payment_state computation,
+        # because the move linked to an expense sheet is of type entry.
+        self.ensure_one()
+        if self.env.context.get('payment_state_matters'):
+            if self.line_ids.expense_id:
```
grindtildeath:
```suggestion
        if self.env.context.get('payment_state_matters') and self.line_ids.expense_id:
```
Nitpicking, but black might be complaining :wink:

hr_expense_sheet_payment_state/models/account_move.py
i-vyshnevska: yes, thanks

i-vyshnevska (approved):

hr_expense_sheet_payment_state/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "HR Expense Report Payment Status",
+    "summary": "Backport of the Payment Status (payment_state field) "
+    "in Expense Report which is introduced in Odoo Community Version 15.0",
+    "version": "14.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
```
dreispt: If this is an Odoo CE backport, shouldn't the author (and copyright) be Odoo SA?

yankinmax: I'm not an expert, so if you definitely know pls suggest the right solution.
@yvaucher can you pls help defining correct license reference?

Saran440: @yankinmax I think you should add Odoo SA. to author and commit with odoo SA.
you can see example with https://github.com/OCA/payroll/pull/5

grindtildeath: @yankinmax Please have a look at the contributing rules on this section: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#10backporting-odoo-modules

Then, as it is a backport of a feature and not a full module, I'd suggest you do the following.
1. Copy the code from Odoo that you will have to work on, and you can commit this with odoo SA, or the original commit's author.
2. Apply your changes on top of it in a second commit that you author yourself.
3. Add requirements from the contributing guidelines.

Saran440:
@yankinmax 
Travis is red because test script error
Could you fix?
Selection_005

yankinmax:
@Saran440 I know the error :smile:
I can't fix it. If you know how pls suggest. I would highly appreciate it.
We don't need to update `number` column in hr.expense.sheet as it doesn't exist.
Maybe some workaround with `@tagged('post_install', '-at_install')`?
I tried and didn't succeed

grindtildeath:
@yankinmax 
Issue comes from module [hr_expense_sequence](https://github.com/OCA/hr-expense/blob/14.0/hr_expense_sequence/models/hr_expense_sheet.py#L12) that defines field `number` as required, therefore installing a DB constraint before the module is loaded to run the tests.
Setting `@tagged('post_install', '-at_install')` should solve the issue.

nimarosa:
@yankinmax please make the changes suggested to get Travis passing and we can move forward with this.

yankinmax:
@nimarosa I've already tried it, setting `post_install....`.
Test doesn't pass. If you have other ideas, can you pls share?

Saran440:
@yankinmax Hello, i think test script error because it not found self.env["ir.sequence"].next_by_code("hr.expense.sheet")
Could you test by add `or "/"` after `number = self.env["ir.sequence"].next_by_code("hr.expense.sheet")`

if it okay we should fix it in hr_expense_cancel and ffw

reference:
 https://github.com/OCA/hr-expense/pull/139/files#diff-d7769905934aa3bf07e8d73608f33b623054d63420886e8026379e7bdf3d4ecaR17

yankinmax:
@Saran440 yep, tested locally, you're right.

yankinmax:
@Saran440 I've updated this PR, including fix for `hr_expense_sequence`:
- [ ] https://github.com/OCA/hr-expense/pull/146

It seems to solve an issue.

yankinmax:
@dreispt this PR is fixed with the fix for `hr_expense_sequence`:
- [ ] https://github.com/OCA/hr-expense/pull/146

Can you pls take a look to get it finally merged?

yankinmax:
@Saran440 I've updated PR and now it can be merged

nimarosa (approved):

PR #84, opened 2021-11-17
[14.0][ADD] hr_expense_journal

Freni-SerpentCS:


dreispt (changes requested):
Thank you Freni.
Please run pre-commit checks, and add the `readme/` subdirectory with the ducmentation fragments.

dreispt (approved):
Fixes added, functional tests made

marcelsavegnago:
@Freni-SerpentCS How about including a field to determine which bank type journals can be selected when posting the expense?

Note: This is just a suggestion and if applicable, it will be implemented in the future.

marcelsavegnago (approved):

dreispt:
@marcelsavegnago Sounds like a good idea, and indeed can be added as an improvement

PR #85, opened 2021-11-18
[FIX] hr_expense_advance_clearing: skip_account_move_synchronization on post journal entries

Saran440:
[BUG] can't post journal entries on expense over advance

Step to error

1. Create advance 100$ and norma process
2. Clear advance 10$
3. Return advance 10$
4. Clear advance **over** 80$
5. Click `Post Journal Entries` it will error.
Selection_004

cc: @kittiu


kittiu:
Thanks for the fix! Can this be FFW?

kittiu (approved):

PR #95, opened 2021-12-21
[14.0][FIX] hr_expense_advance_clearing: default return advance

Saran440:
[BUG]
Return Advance not default from standard
Selection_003

it's because return advance can't use default_get from `account.move`
Reference issue: https://github.com/OCA/hr-expense/issues/44
Reference PR: https://github.com/OCA/hr-expense/pull/47

[FIX]
Call `default_get` from core odoo. it will default standard field.

cc: @kittiu


ps-tubtim (approved):
LGTM.

marcelsavegnago (approved):

kittiu:
Can you try merging again? I don't see any travis error yet.

PR #96, opened 2021-12-22
[14.0][ADD] hr_expense_tax_adjust

ps-tubtim:
This module allows to edit tax amount before Post Journal Entries on Expenses.

**Example**

1. Create an expense with taxes
2. Edit tax amount

tax_adjust


kittiu:
@ps-tubtim the UI of tax amount is not quite inline with other fields, may be you need a bit of adjusting it.

hitrosol:
I try this on my local and it's functionally works. Any plan to reopen and merge this ?

kittiu:
@Saran440 can you open this?

hitrosol:
Thanks for reopening this, few further comments on this.

1. I still can't try this on runboat so I try this only on local.
2. tax tag only can be selected by group_account_readonly, in some cases the plain user is the one who responsible to put the tax tag or not.  It would be good to let the group_user have access on it. 
3. The rounding diff price tax errors also found in this PR, so it will create un balance entry when I try this combine with the hr_expense_advance_clearing module.

Saran440:
@hitrosol Thank you for review this PR.
Can you explain step to error when use with advance clearing module, please?

hitrosol:
Here is the step to error.
1. Install the hr_expense_advance_clearing
2. Set the proper account for product Advance, in this case an Employee Liabilities account with reconcilable
3. Create new Cash Advance, approved, post and register payment.
4. Clear Advance by make expense report with expense item, change the tax amount, submit, approved and try to post the journal, the error shown.
   "Cannot create unbalanced journal entry. Ids: [9]
Differences debit - credit: [1.0]"


https://github.com/OCA/hr-expense/assets/5949936/896423f6-7efa-4583-a0d9-d6d6733837f0

hitrosol:
second thing, is it okay if the field tax can be accessed by internal user ? Because it's easier to change it in the earlier step, instead by group_account_readonly. Especially if we install the hr_expense_tier_validation where some other users already review and approved but later on the Accountant found that the tax need to be adjusted.

Thanks,

hitrosol (approved):
I retry again this in my local with my latest core code.
LGTM. 
Hopefully this can be merged soon. 

Thanks Ecosoft Team

Saran440 (approved):
LGTM :+1:

PR #98, opened 2021-12-28
[14.0][FIX] hr_expense_advance_clearing: return advance - redirect payment message

Saran440:
if you create return advance, it will create and link payment on message
BUT it will go to payment and clear all before path.

This PR change link to redirect payment and you can back to AV is easy.

Before
before

After
after

cc: @kittiu


marcelsavegnago (approved):
LGTM

kittiu (approved):

PR #99, opened 2022-01-17
[14.0][FIX] hr_expense_cancel: Fix tests + Cancel move payments. 

victoralmau:
Fix tests + Cancel move payments (Since odoo/odoo@64b2580).

Please @ernestotejeda and @pedrobaeza can you review it? 

@Tecnativa TT33774

### DIFF
--- a/hr_expense_cancel/models/hr_expense.py
+++ b/hr_expense_cancel/models/hr_expense.py
@@ -107,7 +107,7 @@ class HrExpenseSheet(models.Model):
             account_move = sheet.account_move_id
             sheet.account_move_id = False
             payments = self.env["account.payment"].search(
-                [("expense_sheet_id", "=", sheet.id), ("state", "!=", "cancelled")]
+                [("expense_sheet_id", "=", sheet.id), ("state", "!=", "cancel")]
             )
             # case : cancel invoice from hr_expense
             self._remove_reconcile_hr_invoice(account_move)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Why this is needed? Remember to put always the whys on the commit message and pr description

victoralmau:
Sorry, the PR was in draft until the definitive solution was added. It is already explained a bit and especially the commit that has caused the error in the tests.

ernestotejeda (commented):
Code review 👍🏼

ernestotejeda (approved):

hr_expense_cancel/models/hr_expense.py
```diff
@@ -153,7 +153,7 @@ def _remove_move_reconcile(self, payments, account_move):
 
     def _cancel_payments(self, payments):
         for rec in payments:
-            rec.move_id.with_context({"force_delete": True}).unlink()
+            rec.move_id.button_cancel()
```
pedrobaeza, line 156: Are these payments later reused?

victoralmau: I honestly don't know (I'd have to check it out).

Before the change of odoo with this code the payment was deleted but now it is not deleted and I sincerely believe that it is the right thing to do and that what should be done is to cancel the move (which also cancel the payment).

pedrobaeza (approved):
OK, let's keep it

pedrobaeza:
/ocabot merge patch

Should this go to v15?

victoralmau:
> 

Yes, this will also be necessary in v15 but the migration PR is not yet there.

pedrobaeza:
OK, that's what I meant, if the module is already migrated or not. Check also possible PRs for advertising it.

PR #103, opened 2022-03-03
[15.0][MIG] hr_expense_advance_clearing

kittiu:
Standard migration

Note: I replace clearing one2many field in Advance sheet that show Clearing documents with the use of Smart Button. Because it seem that in v15, the self one2many will show error (one2many of itself like clearings of advance which is same object). Other than that, I also think the use of smart button is always better.


kittiu:
cc @Saran440 @ps-tubtim

kittiu:
@Saran440 @ps-tubtim can you check this?

Saran440:
cherry-pick from https://github.com/OCA/hr-expense/pull/119
and https://github.com/OCA/hr-expense/pull/104

Saran440 (approved):
@ps-tubtim Could you check?

hitrosol:
I still found that if we return advance using Return Button then the amount can't be reconcile in the bank statement.
Other than that, it's works good.

kittiu:
> I still found that if we return advance using Return Button then the amount can't be reconcile in the bank statement. Other than that, it's works good.

I agree, let's explore this time as we do MIG whether we can make it works with Bank Statement too.

hitrosol:
1. I install both this module and account_reconciliation_widget.
2. Create Cash Advance, let's say: IDR 800, submit, approve, post journal entry, register payment.
3. Input bank statement, conduct the bank reconciliation. --> everything is okay.
4. Click Clear Advance button, input the expenses, let's say with total IDR 700, submit to Manager, approve, post journal entries--> works good as well.
5. Since there is IDR 100 left as Cash Advance, then the user click Return advance button.
image

7. It open the Register Payment wizard, then click button Create Payment.
8. Until this step, everything is okay, since the Amount to clear has become IDR 0.
9. Then the accounting would like to reconcile the register payment of the IDR 100 with the actual bank statement
image

Then we see the error message as follow :
image

In V14, i don't use the step number  5. but directly record the receipt in the bank statement instead. 
Then do the journal items reconciliation for Account Employee Advance as related to the Product Advance.

kittiu:
Thanks 👍 @Saran440 @ps-tubtim can you check.

Saran440:
@hitrosol @kittiu 

this commit https://github.com/OCA/hr-expense/pull/103/commits/6a4acea9143d6bac78994ea7e3f8dce95274c95f i add feature to related between payment and advance
Selection_002
Selection_001

if payment is return advance, it will add context `skip_account_move_synchronization` for skip check anymore.
What do you think?

hitrosol:
i try again in my local and it works.... would it be possible to have this in v14 as well ?

Saran440:
@hitrosol i create new pr https://github.com/OCA/hr-expense/pull/136

please check and test

Saran440:
@hitrosol if this pr is okay, Can you help approve this? thank you

ps-tubtim (approved):
LGTM.

PR #104, opened 2022-03-26
[14.0][FIX] hr_expense_advance_clearing: do not allow cancel PV related clearing or return

Saran440:
Payment on advance shouldn't cancel if it related clearing or return advance

Step to test:
1. Create advance 100 $ and normal process until paid (it will create 1 payment)
2. Clearing or return partial advance (20 $) : Clearing Residual = 80 $
3. Go to vendor payment > select payment related from advance (1)
4. reset to draft > cancel **(it shouldn't cancel this payment)**

This PR will check and not allow cancel if clearing residual is not equal amount payment
@kittiu

### DIFF
--- a/hr_expense_advance_clearing/models/__init__.py
+++ b/hr_expense_advance_clearing/models/__init__.py
@@ -3,3 +3,4 @@
 from . import hr_expense
 from . import hr_expense_sheet
 from . import hr_employee_base
+from . import account_move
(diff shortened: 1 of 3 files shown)


Saran440:
@kittiu

kittiu:
@Saran440 I change it to also cover cases clicking "Set Draft", "Cancel" and "Reverse".
Please review / approve.

ps-tubtim (approved):
LGTM.

kittiu (approved):

PR #105, opened 2022-04-09
[14.0][FIX] hr_expense_advance_clearing: not allow return advance if clearing is not posted

Saran440:
Do not allow `return advance` if there are clearing document state `approve`

Example:
1. AV 100
2. Clearing 20 and approve state (not posted yet)
3. Should not `Return Advance` if there are some clearing document is state `approve`
Selection_005

Note: This PR solved conflict budget with `budget_control_advance_clearing`. Should we fix it in this module or `budget_control_advance_clearing`?

----------------------------------------------------
- Small FIX delete payment_mode on onchange

cc: @kittiu


Saran440:
Closed. This PR is not generic on bussiness

PR #109, opened 2022-04-24
[IMP] hr_expense_tier_validation : V14, click submit to manager button before activated the tier validation 

hitrosol:
Before this PR, steps are follows :
1. Save the Expense Report 
2. Request Validation
3. Expense Report validated
4. Click Submit to Manager --> it's weird since the Managers already approved the tier validation.
5. so on and so forth

By this PR:
1. Save the Expense Report
2. Click Submit to Manager  
3. Click Request Validation --> which is better since the requester already submit to manager earlier.
4. Expense Report Validated
6. so on and so forth

### DIFF
--- a/hr_expense_tier_validation/models/hr_expense_sheet.py
+++ b/hr_expense_tier_validation/models/hr_expense_sheet.py
@@ -7,7 +7,7 @@ from odoo import models
 class HrExpenseSheet(models.Model):
     _name = "hr.expense.sheet"
     _inherit = ["hr.expense.sheet", "tier.validation"]
-    _state_from = ["draft"]
-    _state_to = ["submit", "approve", "post", "done"]
+    _state_from = ["submit"]
+    _state_to = ["approve", "post", "done"]
 
     _tier_validation_manual_config = False


hitrosol:
@ps-tubtim, does this  PR make sense for you ?

bunhin:
for the need as on the pr background, this may be achievable by defining the tier definition to only apply to expenses with statuses are in submit stage.

ps-tubtim:
> @ps-tubtim, does this PR make sense for you ?

@hitrosol Sorry for the late reply. IMHO, I think expense reports should be validated by the reviewer before submit to manager because the reviewers should review the expense reports before submitting them to the manager.

hitrosol:
@ps-tubtim, one other benefit of this change is that we can differentiate the expense report  as _draft_ from the _submitted_ one. Draft means that the document hasn't ready to be reviewed and the submitted one means that the document is ready to be reviewed. Currently, the review in tier validation mark as _draft_
We also can put the Manager as part of the Tier Validation, so not nesecarry to the manager has to approve it again since it has been validate in the tier validation.

Saran440:
I agree with @hitrosol because In business, manager should be validate document before send it to external user (tier validation)

FYI I change state tier validation because core odoo change state when create from expense https://github.com/OCA/hr-expense/pull/65

What do you think for this case? @kittiu @ps-tubtim

nimarosa:
@hitrosol is this finished? Please all participants here write your opinion to know if we merge this or it will be automatically closed. 

I agree with the new workflow introduced by this PR, but don't know others using it. Please share your opinions.

hitrosol:
I'm using this in my production db, if others agree, it would be good to merge this. But, I have no authority to do that.

nimarosa:
@hitrosol i Can. That's why I want to see if the others agree. Let's wait a few days to see if other responds.

Saran440:
We use tier validation when in the draft state and skip the submit to manager in production.
I think it depends on the business of each place.

However, this PR is okay. @hitrosol Can you help fix the test script? and let's move on.

Saran440:
@hitrosol @nimarosa Hi, I fix test script and some bug for green travis and move it on https://github.com/OCA/hr-expense/pull/164

nimarosa:
@Saran440 Great. Closed replaced with #164

PR #116, opened 2022-05-25
[14.0][FIX] hr_expense_invoice: do not display UoM and quantity label

LoisRForgeFlow:
Hide quantity label and UoM field when selecting an invoice, as it is done with quantity field.

Before:
Peek 2022-05-25 15-48

After:
Peek 2022-05-25 15-48-2

@ForgeFlow

### DIFF
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -95,6 +95,7 @@ class HrExpense(models.Model):
         """
         if self.invoice_id:
             self.quantity = 1
+            self.product_uom_id = self.product_id.uom_id
             self.tax_ids = [(5,)]
             # Assign this amount after removing taxes for avoiding to raise
             # the constraint _check_expense_ids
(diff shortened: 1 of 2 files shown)


hr_expense_invoice/models/hr_expense.py
```diff
@@ -95,6 +95,7 @@ def _onchange_invoice_id(self):
         """
         if self.invoice_id:
             self.quantity = 1
+            self.product_uom_id = self.product_id.uom_id
```
pedrobaeza, line 98: Why is this needed?

mariadforgeflow (approved):

hr_expense_invoice/models/hr_expense.py
LoisRForgeFlow: To ensure that it is set, maybe the user removed it after setting the product.

pedrobaeza: OK, a very specific case, but let's cover it.

pedrobaeza (approved):

LoisRForgeFlow:
@pedrobaeza merge?

pedrobaeza:
Please fw-port it to v15

PR #119, opened 2022-05-28
[14.0][ENH] hr_expense_advance_clearing, not allow over return

kittiu:
It is ok to make over clearing amount (use more than borrow), but is not making sense to over return amount (return unused amount more than borrow). So this PR prevent that.

### DIFF
--- a/hr_expense_advance_clearing/models/hr_expense_sheet.py
+++ b/hr_expense_advance_clearing/models/hr_expense_sheet.py
@@ -191,3 +191,13 @@ class HrExpenseSheet(models.Model):
             for k, v in clearing_dict.items()
         }
         return clearing_dict
+
+    def action_register_payment(self):
+        action = super().action_register_payment()
+        if self.env.context.get("hr_return_advance"):
+            action["context"].update(
+                {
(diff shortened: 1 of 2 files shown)


kittiu:
cc @Saran440 @ps-tubtim

hr_expense_advance_clearing/models/hr_expense_sheet.py
```diff
@@ -191,3 +191,13 @@ def _prepare_clear_advance(self, line):
...
+
+    def action_register_payment(self):
+        action = super().action_register_payment()
+        if self.advance:
+            action["context"].update(
+                {
+                    "clearing_sheet_ids": self.clearing_sheet_ids.ids,
+                }
+            )
+        return action
```
ps-tubtim: Please fix for multi expenses.

kittiu:
@ps-tubtim fixed.
Note: it is not about multi, because the return advance can't be multi. But I ensure that, the logic only apply for case return advance.

Saran440 (approved):
LGTM :+1:

ps-tubtim (approved):

PR #120, opened 2022-06-08
[14.0][FIX] hr_expense_tier_validation: disallow editable expense from the expense report is under validation

ps-tubtim:


hr_expense_tier_validation/models/hr_expense.py
```diff
@@ -0,0 +1,36 @@
...
+        for val in vals:
+            if val not in exceptions:
+                return False
+        return True
+
+    def write(self, vals):
+        for rec in self:
+            if (
+                rec.sheet_id.state
+                not in ["submit", "approve", "post", "done", "cancel"]
```
kittiu: Do we need this 2 lines?

ps-tubtim: I will change it to `rec.sheet_id.state == "draft"`.

kittiu:
Functional tested and code reviewed. LGTM.

But please add test to cover case,
* When in validation process, do not allow to edit also the hr.expense

kittiu (approved):

Saran440 (approved):
Functional Test :+1:

Saran440:
add test script for cover all case

nimarosa (approved):

PR #122, opened 2022-06-17
[14.0][ADD] Portuguese (Brazilian) Translations

LFPSGS:


nimarosa:
Translation should be added using the translation tools. Refer to https://odoo-community.org/resources/translate for more info.

PR #123, opened 2022-07-04
[ADD][15.0] `sale_expense_cost_reinvoice`

yankinmax:
If expense product standard_price is 0 and expense_policy is `cost` we don't want purchase_price on the order line to be computed to 0.
That is how standard Odoo works
Instead, we will take the Expense price, from just created order line.

With this module if you set Expense product to be re-invoiced at cost you get total amount of the Expense as cost (purchase_price) on order line.


yankinmax:
@OCA/human-resources-maintainers can it be stale?

yankinmax:
@dreispt @alexey-pelykh can we open this PR again? I'll get reviews on it.

yankinmax:
This is superseded by https://github.com/OCA/hr-expense/pull/160

PR #127, opened 2022-09-16
[15.0][MIG] hr_expense_petty_cash

Saran440:
- Delete function `_create_sheet_from_expenses` and add new function `_get_default_expense_sheet_values` for standard expense V15
- Refactored some code for V15


hitrosol:
i like this module since V14, but it would be better if the petty cash holder user group is not group_account_manager but lower access right. Because usually the petty cash holder level is not supposed to see all the manager view, even s/he only see her/his transaction only. Maybe we can start creating new user group for this purpose ?

nimarosa:
I agree with @hitrosol . @Saran440 can you make the changes before merging this?

Saran440:
@nimarosa @hitrosol I change permission from account manager to show full accounting
and move it to accounting menu

Selection_002

nimarosa (approved):

hitrosol:
Thanks for the adjustment of this PR @Saran440 

The creation of new Petty Cash Account and assign it to the group_account_manager or group_account_user is important so not all invoicing user can create and change the max amount, IMO the group_account_manager is much suitable since it's part of the configuration task. Sorry for the confuse with my previous comment.

But we need to add new access right so the model_petty_cash can be accessed by Internal User as readonly who will claim their expense to Petty Cash in Expense Module. By having this, the lower user group, who are not part of the Adminstration group, can manage the transactions of the petty cash. 

Having this changes, also it would be good if the petty cash holder, who are not part of the Adminstration Invoicing, can get the information about the balance of the petty cash easily.

Regards,

hitrosol (approved):
My previous comment is request for the new features. 
But it will not blocked this PR. 
Let's merge this for now.

nimarosa:
@hitrosol please submit a new PR or issue to address your suggestion I think it's important too. 

Merging the migration 

/ocabot merge nobump

PR #132, opened 2022-10-25
[14.0] [ADD] hr_expense_upload_fix

MarcBForgeFlow:
Refresh the product after uploading an expense to trigger all api depends that do not when creating the expense.


MarcBForgeFlow:
@LoisRForgeFlow

PR #169, opened 2023-03-03
[14.0][IMP] hr_expense_invoice: Copy attachments from expense to invoice

victoralmau:
Copy attachments from expense to invoice

@Tecnativa TT41948

### DIFF
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -50,6 +50,11 @@ class HrExpense(models.Model):
                 }
             ]
         )
+        attachments = self.env["ir.attachment"].search(
+            [("res_model", "=", self._name), ("res_id", "in", self.ids)]
+        )
+        for attachment in attachments:
+            attachment.copy({"res_model": invoice._name, "res_id": invoice.id})
         self.write(
             {
                 "invoice_id": invoice.id,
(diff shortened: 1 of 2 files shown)


hr_expense_invoice/models/hr_expense.py
```diff
@@ -24,6 +24,11 @@ class HrExpense(models.Model):
         copy=False,
     )
 
+    def _create_attachments_from_invoice(self, invoice):
```
pedrobaeza: No need of separate method IMO.

hr_expense_invoice/models/hr_expense.py
```diff
@@ -24,6 +24,11 @@ class HrExpense(models.Model):
...
+        res = self.action_get_attachment_view()
```
pedrobaeza: Why do you need this action? Isn't to look for model and ID?

nimarosa:
@victoralmau please Can you rebase this PR to 14.0 branch? So tests will rerun.

victoralmau:
Rebase done.

hr_expense_invoice/models/hr_expense.py
victoralmau: Ok, it was for re-using existing methods, although the domain is very short, it is already changed.

pedrobaeza (approved):

chienandalu (approved):

PR #172, opened 2023-03-14
[15.0][ADD] hr_expense_tax_adjust

Saran440:
This module allows users to edit taxes on the Expense to ensure accurate accounting entries in case of rounding differences. it will accounting journal entries with new price tax that you edit

Peek 2023-03-14 09-49


theangryangel:
Tested and seems functional.

The only nitpick is that it lets you input more tax than the total price, which results in an unclear message when attempting to post the journal entry when using inclusive tax :)

theangryangel (approved):

nimarosa (approved):
Looks useful for some cases. Improving what @theangryangel suggested in the future will be good todo. Merging.

PR #174, opened 2023-03-16
[15.0][FIX] hr_expense_advance_clearing: change field not compute advance

Saran440:
[BUG]
- Duplicate advance sheet, it will change advance to expense

- When create advance sheet and save without line, it will change advance to expense

This PR solved it by change field advance in expense sheet to not compute field


ps-tubtim (approved):
Functional Test: OK.

nimarosa (approved):

PR #176, opened 2023-03-24
[15.0][FIX] hr_expense_advance_clearing: split function return advance and check advance residual

Saran440:
This PR fix 2 commit
- Split function create payment return advance for other module can modify payment before post
- Check advance residual in state submit, approve and post. For case create 2 clearing and clearing first document until advance is no amount left. second clearing must show error


ps-tubtim (approved):
LGTM

TheerayutEncoder (approved):

PR #185, opened 2023-06-07
[14.0][ADD] hr_expense_tags: Add tags to expenses

JoanSoleForgeflow:
cc @ForgeFlow


GuillemCForgeFlow (approved):
LGTM! 👍🏽

hr_expense_tags/models/hr_expense.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class HrExpense(models.Model):
+
+    _inherit = "hr.expense"
+
+    tag_ids = fields.Many2many("crm.tag", string="Tags")
```
marcelsavegnago: Wouldn't it be better to have specific tags for the hr.expense module?

hr_expense_tags/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Hr expense tags",
+    "version": "14.0.1.0.0",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/hr-expense",
+    "license": "AGPL-3",
+    "category": "Human Resources",
+    "depends": ["hr_expense", "crm"],
```
em230418: if "crm" is included because of "crm.tag" usage, then I suggest to use "sales_team" instead.

hr_expense_tags/hooks.py
```diff
@@ -0,0 +1,12 @@
...
+
+def pre_init_hook(cr):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    module = "sales_team"
+    model_data = env["ir.model.data"]
+    for i in range(8):
+        xml_id = module + ".categ_oppor%s" % str(i + 1)
+        record = model_data.xmlid_to_object(xml_id, raise_if_not_found=False)
+        if record:
+            record.unlink()
```
em230418: Why do you delete demo tags?

https://github.com/odoo/odoo/blob/14.0/addons/sales_team/data/crm_tag_demo.xml

hr_expense_tags/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 ForgeFlow - Elmer García / Joan Solé
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Hr expense tags",
```
em230418: How about joining this module and hr_expense_sheet_tags from https://github.com/OCA/hr-expense/pull/186?

hr_expense_tag/models/hr_expense_sheet_tag.py
```diff
@@ -0,0 +1,18 @@
...
+    _description = "Expense sheet Tag"
+
+    def _get_default_color(self):
+        return randint(1, 11)
+
+    name = fields.Char('Tag Name', required=True, translate=True)
+    color = fields.Integer('Color', default=_get_default_color)
+    _sql_constraints = [
+        ('name_uniq', 'unique (name)', "Tag name already exists !"),
+    ]
```
GuillemCForgeFlow, line 18:
```suggestion
```
Maybe we could only add the `hr.expense.tag` model, instead of having two very similar models and then use the model in both the expense and the expense sheet. WDYT?

GuillemCForgeFlow (changes requested):

PR #189, opened 2023-06-21
16.0 mig hr expense advance clearing

jumeldi74:


jumeldi74:
should from 15 to 16. I start from 14 instead.

PR #190, opened 2023-06-25
[15.0][FIX] hr_expense_petty_cash: add permission for internal user

ps-tubtim:
Add permission for internal users to be able to select petty cash holders.


Saran440 (approved):
Code Review

PR #192, opened 2023-07-16
[14.0][FIX] hr_expense_invoice: use bill partner in expense journal item when expense paid by company

JordiBForgeFlow:
If the expense is paid by the company and the expense refers to a bill, then the company will receive charge in the company account from the vendor, not from the employee.

In addition, when the expense is paid by the company the bill should be paid (reconciled), because the expense creates an outstanding payment line, to be reconciled later in the company bank account reconciliation process.

cc @victoralmau @pedrobaeza

### DIFF
--- a/hr_expense_invoice/models/hr_expense_sheet.py
+++ b/hr_expense_invoice/models/hr_expense_sheet.py
@@ -18,12 +18,12 @@ class HrExpenseSheet(models.Model):
         res = super().action_sheet_move_create()
         for sheet in self:
             move_lines = res[sheet.id].line_ids
-            if sheet.payment_mode != "own_account":
-                continue
             for line in self.expense_line_ids.filtered("invoice_id"):
                 c_move_lines = move_lines.filtered(
                     lambda x: x.expense_id == line
                     and x.partner_id == line.invoice_id.commercial_partner_id
+                    and x.account_id.internal_type == "payable"
+                    and not x.reconciled
                 )
                 c_move_lines |= line.invoice_id.line_ids.filtered(
                     lambda x: x.account_id.internal_type == "payable"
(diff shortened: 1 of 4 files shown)


pedrobaeza:
> If the expense is paid by the company and the expense refers to a bill, then the company will receive charge in the company account from the vendor, not from the employee.

Please indicate visually what you refer.

> In addition, when the expense is paid by the company the bill should be paid (reconciled), because the expense creates an outstanding payment line, to be reconciled later in the company bank account reconciliation process.

This module intercepts the move creation, and uses the bill one as target, so I don't understand what you want to get.

victoralmau (commented):
This seems to make sense.

victoralmau (approved):
This seems to make sense.

hr_expense_invoice/tests/test_hr_expense_invoice.py
```diff
@@ -244,8 +244,8 @@ def test_1_hr_test_invoice_paid_by_company(self):
         ).action_sheet_move_create()
         self.assertEqual(self.sheet.state, "done")
         self.assertTrue(self.sheet.account_move_id)
-        # Invoice is not paid
-        self.assertEqual(self.invoice.payment_state, "not_paid")
+        # Invoice is paid
+        self.assertEqual(self.invoice.payment_state, "paid")
```
pedrobaeza, line 248: **issue:** Why this check has been changed?

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -18,12 +18,12 @@ def action_sheet_move_create(self):
         res = super().action_sheet_move_create()
         for sheet in self:
             move_lines = res[sheet.id].line_ids
-            if sheet.payment_mode != "own_account":
-                continue
             for line in self.expense_line_ids.filtered("invoice_id"):
                 c_move_lines = move_lines.filtered(
                     lambda x: x.expense_id == line
                     and x.partner_id == line.invoice_id.commercial_partner_id
+                    and x.account_id.internal_type == "payable"
```
pedrobaeza, line 25: **question:** Why this change?

pedrobaeza (changes requested):

rafaelbn (approved):
Thank you! 👍🏼 

Actually this is an issue since Odoo change in 14 outstanding payment management

@JordiBForgeFlow if you can rebase I could record a video to explain the change in runboat

😄

PR #205, opened 2023-09-18
[16.0][FIX] hr_expense_cancel: Adjust test to product w/o cost behavior

Saran440:


pedrobaeza:
Please explain why this is needed.

Saran440:
@pedrobaeza because demo product is no cost. In standard expense if no cost you can add amount total only.
Selection_011
Selection_012

pedrobaeza:
Weird behavior they have put... Anyway, please add this explanation in the commit message in following lines, and use a better first line commit message:

```
[FIX] hr_expense_cancel: Adjust test to product w/o cost behavior

In this version, if a product doesn't have cost, you are only able
to enter total amount, so the test should be adjusted to this
condition.
```

em230418 (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #207, opened 2023-10-11
[16.0][ADD] hr_expense_analytic_tag: New module

victoralmau:
New module

@Tecnativa TT43223


hr_expense_analytic_tag/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Hr Expense Analytic Tag",
+    "version": "16.0.1.0.0",
+    "category": "Human Resources",
+    "website": "https://github.com/OCA/hr-expense",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["hr_expense", "account_analytic_tag"],
```
pedrobaeza, line 10: auto install?

hr_expense_analytic_tag/models/account_move_line.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2023 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from odoo import models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _prepare_analytic_lines(self):
+        vals = super()._prepare_analytic_lines()
```
pedrobaeza:
```suggestion
        vals_list = super()._prepare_analytic_lines()
```

pedrobaeza: Anyway, this is not the place to do this (apart from the current bad algorithm to match the line). You should put the analytic tags on the `account.move.line`, not on the analytic line, and this will be done automatically.

pedrobaeza (changes requested):

hr_expense_analytic_tag/models/account_move_line.py
victoralmau: Ok, changed to a more appropriate approach.

pedrobaeza (approved):

carlosdauden (approved):
Code LGTM

pedrobaeza:
This should be moved to OCA/account-analytic

PR #212, opened 2023-10-31
[15.0][FIX] hr_expense_tax_adjust: rounding diff price tax

Saran440:
it can error when config Rounding Method is `Round Globally`
This PR is make sure that `diff_price_tax_currency` will compute with rounding.

And support case is not expense (example petty cash)


ps-tubtim (approved):
Functional Test: Ok

PR #225, opened 2024-02-19
[15.0][ADD] hr_expense_employee_analytic_default: New module

victoralmau:
New module

@Tecnativa TT47803


victoralmau:
Ping @pedrobaeza

hr_expense_employee_analytic_default/README.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. contents::
+   :local:
+
+Configuration
+=============
+
+To configure this module, you need to:
+
+#. Go to an employee (or create a new one) and set his private address.
```
pedrobaeza: Let's use the company address instead of the private one for avoiding access errors.

hr_expense_employee_analytic_default/models/hr_expense.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class HrExpense(models.Model):
+    _inherit = "hr.expense"
+
+    @api.onchange("product_id", "date", "account_id")
+    def _onchange_product_id_date_account_id(self):
+        """Set a specific context with the employee of the expense."""
+        _self = self.with_context(hr_expense_employee_id=self.employee_id.id)
+        return super(HrExpense, _self)._onchange_product_id_date_account_id()
```
pedrobaeza: No need to put an extra variable, as `self` is a local one:

```suggestion
        self = self.with_context(hr_expense_employee_id=self.employee_id.id)
        return super()._onchange_product_id_date_account_id()
```

hr_expense_employee_analytic_default/README.rst
victoralmau: I'm sorry but I don't understand. The idea of this module is in fact to be able to set a default analytic account for each employee, if we don't use `address_home_id`, how do we do it?

pedrobaeza: There are 2 linked partners on each `hr.employee`: the private one and the company one. Use the other. As reference, in v17 there's no more `address_home_id`.

victoralmau: Ok, I guess you mean the `user_partner_id` field. It is already changed.

pedrobaeza:
No, sorry for not being specific enough. I mean address_id:

image

That one can be the company one, or the company contact one depending on your needs.

victoralmau:
Changes done.

pedrobaeza (approved):

chienandalu (approved):
Tested :+1:

victoralmau:
This is ready to merge, right?

PR #232, opened 2024-04-30
[16.0][MIG] hr_expense_widget_o2m

Saran440:
For version16, group_ids is not supported in view https://github.com/odoo/odoo/pull/98551
So, I change it to overwrite view when install this module

What do you think? @kittiu


TheerayutEncoder (commented):
LGTM :+1:

PR #234, opened 2024-05-15
[16.0][FIX+IMP] hr_expense_invoice: Proper accounting + UX

pedrobaeza:
- The resulting entries after posting were not correct. Now, this is handled this way:

  * Paid by the company: no entry is created, as the invoice represents the AP balance. The expense sheet payment state is controlled through the linked invoices payment states.
  * Paid by employee: the vendor bill that core creates is voided, and instead, a journal entry transferring the AP balance to the employee is created against the original vendor balance, that is reconciled and given as paid.

- When an invoice is selected in the expense, the interface was not properly adapted, avoiding to change certain data, and the corresponding were not filled as well. Now both things are performed.
- Removed the non-sense register payment constraint.
- Added smart-button to navigate from invoice to expense, as there's no visible link in that direction.
- Avoid the need of the field `with_invoice_id`, as it was simply a question of putting properly the other field as invisible and without groups (apart from the already existing one with groups).
- Tests cleanup, adapted (they were manipulated in some cases to match the weird previous result, like in payment states) and redundancies were removed.

@Tecnativa TT49143


pedrobaeza:
@LoisRForgeFlow can you advice about the error on `hr_expense_tier_validation` ?

```
Traceback (most recent call last):
  File "/__w/hr-expense/hr-expense/hr_expense_tier_validation/tests/test_hr_expense_tier_validation.py", line 84, in test_edit_value_expense
    s.name = "New name"
  File "/opt/odoo/odoo/tests/common.py", line 2278, in __setattr__
    assert not self._get_modifier(field, 'readonly'), \
AssertionError: can't write on readonly field name
```

pedrobaeza:
Checked that the failure is also in main branch: https://github.com/OCA/hr-expense/pull/235

pedrobaeza:
Added the management of payment state and amount residual for expenses paid by employee linked to an invoice with the new transfer entry.

pedrobaeza:
Rebased after the fix in the other module and everything green.

LoisRForgeFlow:
> @LoisRForgeFlow can you advice about the error on `hr_expense_tier_validation` ?
> 
> ```
> Traceback (most recent call last):
>   File "/__w/hr-expense/hr-expense/hr_expense_tier_validation/tests/test_hr_expense_tier_validation.py", line 84, in test_edit_value_expense
>     s.name = "New name"
>   File "/opt/odoo/odoo/tests/common.py", line 2278, in __setattr__
>     assert not self._get_modifier(field, 'readonly'), \
> AssertionError: can't write on readonly field name
> ```

Hi @pedrobaeza 

It seems something specific of the expense tier validation integration, maybe [Ecosoft as the authors](https://github.com/OCA/hr-expense/blob/14.0/hr_expense_tier_validation/__manifest__.py#L8) have a better insight?

pedrobaeza:
Yes, thanks, it has already been fixed in #236. Sorry for not updating here (I just rebased).

LoisRForgeFlow:
> Yes, thanks, it has already been fixed in #236. Sorry for not updating here (I just rebased).

No worries. That was quick :rocket: 

Thanks

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -42,7 +40,35 @@ def _compute_invoice_count(self):
...
-    @api.model
+    @api.depends(
+        "expense_line_ids.invoice_id.payment_state",
+        "expense_line_ids.amount_residual",
+    )
+    def _compute_payment_state(self):
+        """Determine the payment status for lines with expense invoices linked"""
+        invoice_sheets = self.filtered(lambda x: x.expense_line_ids.invoice_id)
+        invoice_sheets.payment_state = "not_paid"
+        for sheet in invoice_sheets:
```
victoralmau, line 51: I think it should be filtered by status so as not to define an incorrect payment state in some cases.

victoralmau (approved):

hr_expense_invoice/models/hr_expense_sheet.py
pedrobaeza: They are already filtered in the first line of the method, but all the expenses with a linked invoice should be computed in this case.

carlosdauden (approved):

sergio-teruel (approved):

SoniaViciana:
It's not OK this merge. Now, if this case occurs:

image

The accounting entries are wrong:
* FACTU/2024/05/0003: Contains the journal line to an expense account (629)
* When I "Post Journal Entries", now the expense account it's duplicated, this is the result:

image


Before this merge, **It was against a payable account (410)**, that would be the right thing to avoid having duplicate expenses.

Also, it's wrong in this other case:

image

If I "Cancel all related operations", the payment made to the invoice will not be canceled. And i have another button ("Register Payment") that right now it's of no use.


Can you review it please? @pedrobaeza 
Thanks in advance,

pedrobaeza:
Please open an issue, but that was not the behavior got, as you can check in the tests. Please do the tests on a runboat and report back in an issue.

gonzalolenzi:
Hi @pedrobaeza , we're having some problems using this module after this specific change.

I'm attaching a video (00:59 seconds) from runboat without any change in the configuration just so you understand the flow I'm trying to follow and the problem I'm having

https://drive.google.com/file/d/1celILhfky3xYFZw1WsYM3f0isrMIm1iu/view

The flow we were doing before this change was:
1. Create the expense
2. Create the report
3. Create the vendor bill and validate it.
4. Post journal entries and generate employee debt
5. Register payment 

I think with the change, the flow should be the same because we are generating the AP through the entry in the general journal, but for some reason I can not register the payment. I get "You can't register a payment because there is nothing left to pay on the selected journal items." like there is no debt related.

Can you help me? 

Thanks in advance.

pedrobaeza:
@gonzalolenzi please check #249. There's still a couple of things left that I have to complete though.

Issue #248, opened 2024-06-18
[16.0][BUG] hr_expense_invoice: Cannot Edit Vendor Bill Created from Multiple Expenses


rafaelbn:
### Bug: Cannot Edit Vendor Bill Created from Multiple Expenses

I have encountered an issue with the hr_expense_invoice module in Odoo, specifically when trying to create a vendor bill from multiple expenses. When the total expense amount is €115 and the "Create Vendor Bill" button is used to generate the invoice, the resulting vendor bill cannot be edited to properly account for the expenses.
Steps to Reproduce:

- Go to an expense report with multiple expenses totaling €115. **Don't add any tax manually! 😄** 
- Click the "Create Vendor Bill" button.
- Observe that the created vendor bill cannot be edited propertly -> **video 2 minutes 1x, 1 minutes 2x**

   https://www.loom.com/share/dd5827e3969c4fefbfa127008f48dea7?sid=5026f96a-5ca8-41ce-bed0-b91a0aeb6258

 
**Expected Behavior:**

The vendor bill created from multiple expenses should be editable, allowing for proper accounting and adjustments as needed.

**Actual Behavior:**

The created vendor bill is not editable, preventing necessary adjustments and correct accounting of the expenses.

Thank you very much for you attention
Best regards,
Rafael

@moduon MT-5202 

Please could you give an opinion @SoniaViciana @victoralmau @kwoychesko @JordiBForgeFlow @alexey-pelykh @dreispt @tafaRU @gurneyalex ?

Thank you!!!! 😄 ❤️ 
Best regards!
Rafael

References:
- https://github.com/OCA/hr-expense/issues/245 (Issue closed in 2 minutes - Fastest ever - Promotes contributions and really compliance with OCA rules about how to welcome new-contributors)
- https://github.com/OCA/hr-expense/pull/244 (PR closed in 18 minutes)

rafaelbn:
Solved in https://github.com/OCA/hr-expense/pull/268

PR #249, opened 2024-06-22
[16.0][FIX] hr_expense_invoice: Incorrect data and check when creating vendor bill from expense

### LINKED ISSUE #245: [BUG] hr_expense_invoice: errors when a tax is set in the expense.
There is an error in the expense sheet when we have set the taxes. Also if we try to create the expense sheet without taxes and then correct it in the invoice we will get the error that the total of the invoice does not match the total of the expense.

At the moment we create an invoice from the expense sheet the information of the tax fields is deleted, leaving the records incoherent.

In the following PR I propose a solution: https://github.com/OCA/hr-expense/pull/244

Here is a video explaining the issue: https://www.loom.com/share/a6c6049e0fbf431180a6b904e5694b13

## Module
hr_expense_invoice

## To Reproduce
**Affected versions**: 16.0 but probably more.

1. Create an expense, set up a total and a tax.
2. Click on "CREATE REPORT" to create the expense sheet. Then click on the expense sheet line "Create Vendor Bill".
Make sure the tax has propagated to the expense line.

You will see how the expense total changed and the invoice as well. 

**Expected behavior**
The total invoiced and the total expense should be the same.

pedrobaeza:
Two fixes:

1. The "amount_total" field now reflects the total of the invoice including taxes, so we have to use "untaxed_amount".
2. If changing any data in the linked vendor bill, the computed fields used for the comparison are not yet recomputed, so let's do the check on a later stage (after the write).

Fixes #245

@Tecnativa

### DIFF
--- a/hr_expense_invoice/models/account_move.py
+++ b/hr_expense_invoice/models/account_move.py
@@ -19,8 +19,13 @@ class AccountMove(models.Model):
         "transfer journal entry",
     )
 
-    @api.constrains("amount_total")
-    def _check_expense_ids(self):
+    def write(self, vals):
+        # Check if the amount of the invoice linked to an invoice is different
+        # Done here in the write instead of a Python constraint as the computed field
+        # amount_total is not yet updated on that moment
+        res = super().write(vals)
+        if self.env.context.get("skip_account_move_synchronization"):
+            return res
         DecimalPrecision = self.env["decimal.precision"]
         precision = DecimalPrecision.precision_get("Product Price")
         for move in self.filtered("expense_ids"):
@@ -32,6 +37,7 @@ class AccountMove(models.Model):
                         "linked to this invoice."
                     )
                 )
+        return res
 
     def action_view_expense(self):
         self.ensure_one()
--- a/hr_expense_invoice/models/hr_expense.py
+++ b/hr_expense_invoice/models/hr_expense.py
@@ -36,7 +36,7 @@ class HrExpense(models.Model):
                 {
                     "product_id": self.product_id.id,
                     "name": self.name,
-                    "price_unit": self.unit_amount or self.total_amount,
+                    "price_unit": self.untaxed_amount,
                     "quantity": self.quantity,
                     "account_id": self.account_id.id,
                     "analytic_distribution": self.analytic_distribution,
@@ -126,6 +126,8 @@ class HrExpense(models.Model):
         )
         for attachment in attachments:
             attachment.copy({"res_model": invoice._name, "res_id": invoice.id})
+        # Normalize data in the expense for avoiding mismatchings. Examples: the tax
+        # reported is not correct, there's more than one concept, etc...
         self.write(
             {
                 "invoice_id": invoice.id,


edlopen:
Hello @pedrobaeza , can you reset de runboat so we can test it? Thank you!

pedrobaeza:
Re-pushed

rafaelbn:
FIX : https://github.com/OCA/hr-expense/issues/248

@moduon MT-5202 Please @edlopen @Gelojr could you review 😄 ❤️

edlopen (approved):
LGTM. Code and functional review. Thank you @pedrobaeza.

edlopen (changes requested):
Hello @pedrobaeza , I've came across a new issue with the flow you suggested.
If in the bill I add a Vendor or a reference the validation error show up again.

https://www.loom.com/share/989658911ea94a788dc820adee39b5dd?sid=1b33c2b6-16d8-4ded-8613-e7d680d25761

loida-vm (changes requested):
Hello, upon reviewing the PR, I don't see that it follows the correct flow. If the expense is X, the total amount of the invoice should also be X. Therefore, if we add a tax to the invoice, the price must change so that the total remains X.

When I perform these steps, a warning appears and does not allow us to confirm the invoice because it is based on the taxable base and not on the total. This is incorrect. 

You can see the flow in the following video: https://www.loom.com/share/9c620effe6174653b380aa7a09d403ae?sid=631e7f66-ca85-42e8-8d54-d4cbb7450a37

Shide:
@pedrobaeza can I continue this PR in a separate PR using your code (and mentioning you)?

pedrobaeza:
We can merge this one and you add an extra PR on top of it. These fixes are welcome although not completed.

hr_expense_invoice/models/account_move.py
```diff
@@ -19,8 +19,11 @@ class AccountMove(models.Model):
         "transfer journal entry",
     )
 
-    @api.constrains("amount_total")
-    def _check_expense_ids(self):
+    def write(self, vals):
+        # Check if the amount of the invoice linked to an invoice is different
+        # Done here in the write instead of a Python constraint as the computed field
+        # amount_total is not yet updated on that moment
+        res = super().write(vals)
```
Shide, line 26:
```suggestion
        res = super().write(vals)
        if self.env.context.get("skip_account_move_synchronization"):
            return res
```

Shide (changes requested):
I've spotted an strange behaviour.
If you perform a `write` when you are changing the supplier and the lines (always with a match on the total amounts) at the same time (modify all the fields you need to change and click save), the `move.amount_total` value is the old one.

The key to know when to allow Odoo perform it's changes or checks is to add the context checking.

With this small modifications works perfect.

hr_expense_invoice/models/account_move.py
Shide: @pedrobaeza can you address it? 

@loida-vm  @edlopen please approve this PR in order to allow me to continue doing things

Shide (approved):

hr_expense_invoice/models/account_move.py
pedrobaeza: Changed, @Shide

edlopen (approved):
LGTM indeed!

loida-vm (commented):
I’m presenting a video here doing the functional check.
As you can see, and as my colleague @Shide  mentioned, the value move.amount_total it "remembers" is the old one. Therefore, if the case is like the one I’m presenting (the second) in the video, it still doesn't allow me to confirm the invoice.
https://www.loom.com/share/23ba25ca6c664f42af38a3ae70b09b20?sid=145650c5-cc3c-4e07-bd78-73a03646ca3b

pedrobaeza:
@loida-vm was that test done before or after I performed the change that @Shide suggested?

loida-vm:
It was this morning, after the change

pedrobaeza:
Well, then the patch suggested by @Shide seems to be not working in that case.

Shide:
@loida-vm I'll fix your second case soon.

loida-vm:
Ok, thank you very much. Aside from that case, the PR works correctly.

ProcessControl-JHF:
Will we have this resolved in version 17? Thanks

PR #259, opened 2024-08-02
[16.0][FIX] hr_expense_advance_clearing: visible analytic

Saran440:
This PR delete code invisible analytic.
I can't remember why we invisible analytic in advance?

What do you think? @kittiu

### DIFF
--- a/hr_expense_advance_clearing/views/hr_expense_views.xml
+++ b/hr_expense_advance_clearing/views/hr_expense_views.xml
@@ -141,11 +141,6 @@
                     placeholder="Optional clearing product is used during clear advance"
                 />
             </field>
-            <field name="analytic_distribution" position="attributes">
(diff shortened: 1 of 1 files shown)


douglascstd:
Hi @Saran440, 

When using advances, we cannot debit or credit an analytic account because the advance is not considered an expense. If we debit an analytic account for the advance and later use it to offset an expense with an analytic account, it results in two debits. Therefore, the analytic account should be nullified.

There is an error (possibly needing another PR to address): If the analytic account field is filled and then changed to an advance, the content of the field remains.

Saran440:
@douglascstd Thank you for the explanation.
I will move it back to customize module.

and the issue you mentioned, if it still error please open and add step here https://github.com/OCA/hr-expense/issues

PR #270, opened 2024-09-27
[17.0][MIG] hr_expense_employee_analytic_default: Migration to 17.0

victoralmau:
Migration to 17.0

@Tecnativa TT50068


chienandalu (commented):
Maybe the module should be renamed to something like hr_expense_employee_analytic_distribution?

victoralmau:
I agree that the current name is not the right one, maybe `hr_expense_employee_analytic_distribution_default`?

What do you think @pedrobaeza ?

pedrobaeza (approved):
I think it's OK to keep the name, as there's no better one.

/ocabot migration hr_expense_employee_analytic_default

chienandalu (approved):

Issue #273, opened 2024-10-18
[16.0 & 17.0]  hr_expense_invoice error when expense report includes invoices from diferent providers


TheCloneMaster:
## Module
hr_expense_invoice

## Describe the bug
If an expense report includes invoices from diferent providers, when posting move the report moves to "paid" and invoices are not reconciled as paid.

## To Reproduce
**Affected versions**: 16.0  & 17.0

1. Create vendor bill for "provider 1"
2. Create vendor bill for "provider 2"
3. Create expense report, one line for bill created for provider 1 and another line for provider 2
4. Approve and post expense report
As a result, the expense report is marked as paid, and the invoices remain open

**Expected behavior**
The Expense report should move to "register payment", and the original invoices should appear as "paid"

TheCloneMaster:
The error is in _do_create_moves on hr_expense_sheet.py.
line 40 asks for the wrong vendor id...
                    == self.expense_line_ids.invoice_id.partner_id

it should be 
                    == expense.invoice_id.partner_id

so that it takes the partner_id of current expense, not the one from the first expense from the report

I will upload a PR  

Regards,
-Mario

PR #274, opened 2024-10-21
[Fix 17.0] hr_expense_invoice

TheCloneMaster:
Fixes issue:
https://github.com/OCA/hr-expense/issues/273

If an expense report includes invoices from diferent providers, when posting move the report moves to "paid" and invoices are not reconciled as paid, as it fails to reconcile the lines with the transfer moves.

Use the correct partner, from current expense line from the report, not always the same partner from first expense line

PR #275, opened 2024-10-21
[Fix 16.0] hr_expense_invoice

TheCloneMaster:
Fixes issue:
https://github.com/OCA/hr-expense/issues/273

If an expense report includes invoices from diferent providers, when posting move the report moves to "paid" and invoices are not reconciled as paid, as it fails to reconcile the lines with the transfer moves.

Use the correct partner, from current expense line from the report, not always the same partner from first expense line


pedrobaeza:
The same comments as in #274 

If you use GitHub for editing the file, please do it with the correct message format.

TheCloneMaster:
Hope it has correct format now
Thanks

TheCloneMaster:
I will close it and apply again, following your guidelines
Thanks!

PR #281, opened 2024-12-06
[18.0][MIG] hr_expense_cancel

mav-adhoc:


baf-adhoc (approved):
Functional review ok

ced-adhoc (approved):
LGTM

ced-adhoc:
Hi @Saran440 , could you please check this pr? thanks!

hr_expense_cancel/models/hr_expense.py
```diff
@@ -0,0 +1,66 @@
...
+            # Deleting the Journal entry if in the previous steps
+            # (if the expense sheet is paid and payment_mode == 'own_account')
+            # it has not been deleted
+            if account_moves.exists():
+                moves_to_cancel = account_moves.filtered(
+                    lambda move: move.state != "draft"
+                )
+                if moves_to_cancel:
+                    moves_to_cancel.button_cancel()
+                account_moves.with_context(force_delete=True).unlink()
```
Saran440, line 31:
```suggestion
                account_moves.with_context(force_delete=True).unlink()
```

In my opinion, some accountants may not be comfortable with deleting documents. Instead, the move should be reversed for audit purposes.
Should this line include a new method as a hook?

mav-adhoc: Maybe we should include this improve after the migration. What do you think?

PR #283, opened 2024-12-27
[18.0][MIG] hr_expense_employee_analytic_default: Migration to 18.0 

BhaveshHeliconia:


marcelsavegnago (approved):
LGTM

CRogos (commented):
Code LGTM.


> 3.  When changing the employee, the analytical distribution will
>     be updated.

When changing to a product, where no analytic distribution exists, the analytic accounts are not removed but stay with the previous distribution.

I've also tried to set a distribution for Partner "Anita Oliver", but when I select the employee "Anita Oliver" the distribution is not applied.

CRogos:
@BhaveshHeliconia same findings like before. Did you change anything or just a rebase?

BhaveshHeliconia:
@CRogos : Earlier I had only rebased, but now I’ve also updated the README file.

CRogos:
I agree with the readme change, but changing the employee did not work, and also the analytic distribution did not get removed, when I change to a product without distribution template.

BhaveshHeliconia:
> I agree with the readme change, but changing the employee did not work, and also the analytic distribution did not get removed, when I change to a product without distribution template.

Okay, I’ll work on fixing that.

BhaveshHeliconia:
@CRogos : The module now updates and clears the analytic distribution as expected when changing the employee or product, both interactively and on save. I have also added test cases for the onchange logic.
Could you please review it when you get a chance? Thanks in advance.

CristianoMafraJunior (approved):
LGTM

CRogos (commented):
@BhaveshHeliconia we are currently not using this module. Therefore I am not really into it and my first test failed, most likely because I did not set the work_contact_id correct.

BhaveshHeliconia:
@victoralmau could you please review this module when you get a chance? Thank you in advance!

hr_expense_employee_analytic_default/models/account_analytic_distribution_model.py
```diff
@@ -15,6 +15,6 @@ def _get_distribution(self, vals):
             employee = self.env["hr.employee"].browse(
                 self.env.context.get("hr_expense_employee_id")
             )
-            if employee.address_id:
```
victoralmau: I don't understand why this field was changed and a different one is now being used; in my opinion, it would be better to use `address_id`.

hr_expense_employee_analytic_default/models/hr_expense.py
```diff
@@ -7,6 +7,28 @@
...
+                "company_id": self.company_id.id,
+            }
+            distribution = (
+                self.env["account.analytic.distribution.model"]
+                .with_context(hr_expense_employee_id=self.employee_id.id)
+                ._get_distribution(vals)
+            )
+            self.analytic_distribution = distribution or {}
+        else:
+            self.analytic_distribution = {}
```
victoralmau: In my opinion, this `onchange` method is unnecessary; the `_compute_analytic_distribution()` method already adds the appropriate context `hr_expense_employee_id` to define the corresponding analytical distribution. For this reason, the `_get_distribution()` method uses the appropriate context value to perform the necessary actions there.

victoralmau (changes requested):

hr_expense_employee_analytic_default/models/account_analytic_distribution_model.py
BhaveshHeliconia: Actually address_id is the company work address — it's computed from company_id.partner_id, so all employees in the same company end up with the same address_id. That wouldn't work here since we need to match distribution rules per employee.

If you check Odoo core's _compute_analytic_distribution() in hr_expense.py (L522-523), it also uses work_contact_id for the partner matching. So we're staying consistent with how core handles it.

hr_expense_employee_analytic_default/models/hr_expense.py
BhaveshHeliconia: you're right. The _compute_analytic_distribution() already depends on employee_id and product_id, and the context override we add takes care of the rest through _get_distribution(). The onchange is just duplicating that logic. I'll remove it and update the tests.

victoralmau (changes requested):
Reviewing hr_expense in v18 https://github.com/odoo/odoo/blob/ffba9eb97876c2507f3bc137298388bf96db6b16/addons/hr_expense/models/hr_expense.py#L522, the appropriate `partner_id` (the employee's) is already defined; therefore, this module would no longer be necessary because `hr_expense` already does the same thing.

BhaveshHeliconia:
> Reviewing hr_expense in v18 https://github.com/odoo/odoo/blob/ffba9eb97876c2507f3bc137298388bf96db6b16/addons/hr_expense/models/hr_expense. py#L522, the appropriate `partner_id` (the employee's) is already defined; therefore, this module would no longer be necessary because `hr_expense` already does the same thing.

You're right, I checked and in v16 the _compute_analytic_distribution didn't consider the employee at all — no partner_id or employee_id in the depends or vals. That's why this module was needed back then. But in v18 they added employee_id as a dependency and pass work_contact_id as the partner_id directly, so the core now does exactly what this module was doing. It looks like this module is no longer needed for v18.

MiquelRForgeFlow:
Then please close the PR.

PR #298, opened 2025-04-18
[17.0][MIG] hr_expense_journal

Pani-k-folk:
**change**

Remove the field payment_type_id because in version 17.0, the selection has changed from choosing the journal directly to choosing based on the payment method.

Add the fields payment_method_line_id and selectable_payment_method_line_ids to allow selecting the payment method based on the journal.


yboujraf (approved):
Dear @Saran440 , @Pani-k-folk 

I tested and it works as expected.

image

image

Best Regards,

TheerayutEncoder (approved):
LGTM :+1:

PR #305, opened 2025-06-16
[15.0][IMP] hr_expense_sequence_option: change ir.sequence.option from demo to data

Pani-k-folk:
Change ir.sequence.option from demo to data ,so that it can be referenced by XML.


ps-tubtim (approved):

Saran440 (approved):
LGTM 👍

PR #308, opened 2025-07-03
[18.0][MIG] hr_expense_invoice: Migration to version 18.0

david-banon-tecnativa:
@tecnativa @pedrobaeza @victoralmau 
Superseedes https://github.com/OCA/hr-expense/pull/282 by @cav-adhoc while maintaining commit ownership.

I've made some minor adjustments, orignal PR looked good to me and only needed a rebase to work.

TT55529

PR #309, opened 2025-07-07
[18.0][FIX]  hr_expense_advance_clearing: Check approval before other modules might intervene.

david-banon-tecnativa:
While migrating hr_expense_invoice https://github.com/OCA/hr-expense/pull/308, it conflicted with this module, so i made a fix.

This is the error in ci: https://github.com/OCA/hr-expense/actions/runs/16047191168/job/45281217754#step:8:176

I don't have enough experience with this module to know if it will have undesired side effects, however in my tests, i haven't found any.

I also fixed translations for validation errors, as f strings are not properly translatable.

Can you please review?

@kittiu @Saran440 

@Tecnativa TT55529


pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

PR #311, opened 2025-08-04
[17.0][IMP] hr_expense_tier_validation: Change the use of the validated field to validation_status

victoralmau:
FWP from 16.0: https://github.com/OCA/hr-expense/pull/310

Change the use of the `validated` field to `validation_status`

Related to https://github.com/OCA/server-ux/pull/1084

Locked by:
- [ ] `base_tier_validation`: https://github.com/OCA/server-ux/pull/1136

Please @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/hr_expense_tier_validation/models/hr_expense.py
+++ b/hr_expense_tier_validation/models/hr_expense.py
@@ -40,8 +40,8 @@ class HrExpense(models.Model):
             if (
                 sheet.state == "submit"
                 and sheet.review_ids
-                and not sheet.validated
-                and not sheet.rejected
+                and not sheet.validation_status == "validated"
+                and not sheet.validation_status == "rejected"
                 and not rec._check_allow_write_under_validation(vals)
             ):
                 raise ValidationError(_("The expense report is under validation."))
--- a/hr_expense_tier_validation/tests/test_hr_expense_tier_validation.py
+++ b/hr_expense_tier_validation/tests/test_hr_expense_tier_validation.py
@@ -63,7 +63,6 @@ class TestHrExpenseTierValidation(TestExpenseCommon):
             sheet.action_approve_expense_sheets()
         sheet.request_validation()
         self.assertTrue(sheet)
-        sheet.invalidate_model()
         # tier validation but state still submit
         self.assertEqual(sheet.state, "submit")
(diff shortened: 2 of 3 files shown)


pedrobaeza:
Maybe we should preserve the fields as computed from the `validation_status` to avoid all of these PRs?

victoralmau:
Seriously? They've already been removed from v16 (and all PRS merged) and all PRs for v17 have been done, and now we're considering that maybe we should have left them in? I specifically mentioned that I preferred to leave them in https://github.com/OCA/server-ux/pull/1084#discussion_r2225924560 but the decision was made to remove them.

pedrobaeza:
Yeah, the objective is to not have 2 sources of truth, but maybe remove them at all instead of converting them into stored fields was too aggressive (my fault).

pedrobaeza:
@StefanRijnhart should we re-add the fields in 16 and keep them in 17 as computed? What do you think?

StefanRijnhart:
@pedrobaeza Yes please. That was our original approach in https://github.com/OCA/server-ux/pull/1101: trivial compute and search methods based on validation_status.

StefanRijnhart:
@pedrobaeza Maybe consider them in 18.0 as well, which has also matured quite a bit by now. And we'll all vow to prevent them from getting through in 19.0.

pedrobaeza:
OK, sorry for not thinking first in this. Having them as computed non stored with search method is a middle ground, or should we go with stored as well?

StefanRijnhart:
@pedrobaeza I would not store, as it violates single source of truth like you argued.

victoralmau:
Finally, the `validated` and `rejected` fields remain unchanged (as before) https://github.com/OCA/server-ux/pull/1136. 
Do you think that this change (and similar changes in other PRs for other modules) should be done to avoid any possible confusion? To continue with all those that are open or close them.

pedrobaeza:
Well, if the code in this module is compatible with no change (which is the desirable thing), we shouldn't perform any change IMO.

victoralmau:
> Well, if the code in this module is compatible with no change (which is the desirable thing), we shouldn't perform any change IMO.

Okay, clarified, I'm closing all related PRs.

PR #312, opened 2025-08-29
[18.0][MIG] hr_expense_widget_o2m

TheerayutEncoder:
Standard Migration

Hi @kittiu, @Saran440 
Relate to 

- https://github.com/OCA/hr-expense/pull/232 
- https://github.com/odoo/odoo/pull/98551 
- https://github.com/odoo/odoo/pull/95729 

on version 18.0 groups_id also removed from ir.ui.view 
That means we cannot configure it to use O2M widget by groups; we can only enable this feature by installing or uninstalling it.


Saran440 (approved):
Code Review :+1:

Saran440 (changes requested):
Step to test:
1. Create hr.expense.sheet and add a line
2. Change quantity or Total > Save
3. Click in each line > edit price > Save
4. Click from smart button expense

Result:
The price still shows the old value.

I think this behavior may be related to the widget implementation in Odoo 18 and the way the x2many state is handled.

To simplify the user experience, I created an alternative module using the standard Odoo behavior with `open_form_view`:
https://github.com/OCA/hr-expense/pull/342

What do you think?

TheerayutEncoder:
> Step to test:
> 
> 1. Create hr.expense.sheet and add a line
> 2. Change quantity or Total > Save
> 3. Click in each line > edit price > Save
> 4. Click from smart button expense
> 
> Result: The price still shows the old value.
> 
> I think this behavior may be related to the widget implementation in Odoo 18 and the way the x2many state is handled.
> 
> To simplify the user experience, I created an alternative module using the standard Odoo behavior with `open_form_view`: #342
> 
> What do you think?

Yes, it's better to just use standard odoo like `open_form_view`.

PR #319, opened 2025-11-17
[18.0][FIX] hr_expense_invoice: Handle mixed lines

christian-ramos-tecnativa:
Before this PR, if an expense report contained a mix of lines with invoices and lines without invoices, the AP transfer was not created.
A small adjustment was made to `test_5`, and a new test was added to explicitly cover this scenario (even though `test_5` now also validates it).

@Tecnativa TT59171

### DIFF
--- a/hr_expense_invoice/models/hr_expense_sheet.py
+++ b/hr_expense_invoice/models/hr_expense_sheet.py
@@ -17,18 +17,9 @@ class HrExpenseSheet(models.Model):
             lambda sheet: func(expense.invoice_id for expense in sheet.expense_line_ids)
         )
 
-    def _do_create_moves(self):
-        """Don't let super to create any move:
-        - Paid by company: there's already the invoice.
-        - Paid by employee: we create here a journal entry transferring the AP
-          balance from the invoice partner to the employee.
-        """
-        expense_sheets_with_invoices = self.get_expense_sheets_with_invoices(all)
-        res = super(
-            HrExpenseSheet, self - expense_sheets_with_invoices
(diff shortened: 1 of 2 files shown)


hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -17,19 +17,12 @@ def get_expense_sheets_with_invoices(self, func):
...
-        res = super(
-            HrExpenseSheet, self - expense_sheets_with_invoices
-        )._do_create_moves()
+    def _do_create_ap_moves(self):
         # Create AP transfer entry for expenses paid by employees
-        for expense in expense_sheets_with_invoices.expense_line_ids:
-            if expense.payment_mode == "own_account":
+        for expense in self.expense_line_ids.filtered(
+            lambda expense: expense.invoice_id
+        ):
```
pedrobaeza: Use the available syntactic sugar:

```suggestion
        for expense in self.expense_line_ids.filtered("invoice_id"):
```

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -17,19 +17,12 @@ def get_expense_sheets_with_invoices(self, func):
...
+            if expense.invoice_id and expense.payment_mode == "own_account":
```
pedrobaeza: You have already check `invoice_id` before.

pedrobaeza (approved):

david-banon-tecnativa (approved):

PR #331, opened 2026-03-09
[18.0][FIX] hr_expense_invoice: Allow mixed expense reports (with and without invoices)

### LINKED ISSUE #330: [FIX][18.0] hr_expense_invoice: Allow mixed expense reports (with and without invoices)
In v18, hr_expense_invoice does not allow unifying expenses without an associated invoice (e.g., Expense 1) and expenses with an associated vendor invoice (e.g., Expense 2) in the same report, both paid by the company. This forces the creation of a separate report for each type of expense. When attempting to approve a single report containing both types of expenses, an "Invalid Operation" error occurs. I have attached images.

Could this be fixed with a PR? This is a problem for our client, but would it make sense for the OCA?

Image

Image

Image

Image

Thank you in avance
Best regards

@pedrobaeza @victoralmau @kittiu

dtec-landoo:
**Description of the issue/feature:**

Fixes an issue where expense reports containing a mix of expenses with an associated vendor bill and expenses without a vendor bill could not be processed, raising an "Invalid Operation" error.

Currently, the hr_expense_invoice module assumes that either all or none of the expenses in a report have an associated invoice. When a mixed report is submitted, the _check_can_create_move constraint and the _do_create_moves method fail to handle the non-invoiced lines properly, crashing the approval/posting process.

This PR fixes the issue by:

    Modifying _do_create_moves to temporarily remove invoiced expenses from the sheet before calling super(). This allows Odoo's standard process to generate the regular journal entries for non-invoiced expenses. Afterwards, the invoiced expenses are restored to the sheet and their specific AP transfer moves are generated.

    Changing any to all in the _check_can_create_move validations so mixed sheets are not incorrectly blocked.

    Modifying action_sheet_move_post to properly filter sheets passing to super().

    Bonus fix: Adds support for in_refund moves (vendor refunds) in _check_expense_ids so negative expenses don't raise a validation error when comparing amount_total.

**Steps to reproduce:**

    Install hr_expense_invoice.

    Create an Expense without a vendor bill (e.g., "Expense 1", Paid by Company).

    Create an Expense with an associated vendor bill (e.g., "Expense 2", Paid by Company).

    Select both expenses and create a single Expense Report.

    Submit the report and click on "Approve" / "Post Journal Entries".

**Current behavior:**

An "Invalid Operation" (or validation) error is raised because the system tries to process the mixed lines uniformly, failing to generate the standard accounting entries for the non-invoiced expenses. This forces the user to split the expenses into separate reports.

**Expected behavior:**

The mixed expense report is successfully approved and posted. Non-invoiced expenses generate their standard journal entries, and invoiced expenses are processed correctly alongside them within the same report.

Fixes #330

### DIFF
--- a/hr_expense_invoice/models/hr_expense_sheet.py
+++ b/hr_expense_invoice/models/hr_expense_sheet.py
@@ -19,37 +19,81 @@ class HrExpenseSheet(models.Model):
 
     def _do_create_ap_moves(self):
         # Create AP transfer entry for expenses paid by employees
+        all_generated_moves = self.env["account.move"]
         for expense in self.expense_line_ids.filtered("invoice_id"):
             if expense.payment_mode == "own_account":
                 move_vals = expense._prepare_own_account_transfer_move_vals()
                 move = self.env["account.move"].create(move_vals)
-                move.action_post()
-                # reconcile with the invoice
-                ap_lines = expense.invoice_id.line_ids.filtered(
-                    lambda x: x.display_type == "payment_term"
+                all_generated_moves |= move
+        return all_generated_moves
+
+    def _reconcile_ap_moves(self):
+        # Reconcile AP transfer entries with vendor bills once they are posted
+        for expense in self.expense_line_ids.filtered("invoice_id"):
+            if expense.payment_mode == "own_account":
+                move = self.account_move_ids.filtered(
+                    lambda m, exp=expense: m.source_invoice_expense_id == exp
                 )
-                transfer_line = move.line_ids.filtered(
-                    lambda x, partner=expense.invoice_id.partner_id: x.partner_id
-                    == partner
-                )
-                (ap_lines + transfer_line).reconcile()
+                if move and move.state == "posted":
+                    ap_lines = expense.invoice_id.line_ids.filtered(
+                        lambda x: x.display_type == "payment_term"
+                    )
+                    transfer_line = move.line_ids.filtered(
+                        lambda x, partner=expense.invoice_id.partner_id: x.partner_id
+                        == partner
+                    )
+                    if ap_lines and transfer_line:
+                        (ap_lines + transfer_line).reconcile()
+
+    def _do_create_moves(self):
+        removed_expenses_map = {}
+        sheets_to_skip_super = self.env["hr.expense.sheet"]
+        all_generated_moves = self.env["account.move"]
+        for sheet in self:
+            expenses_with_invoice = sheet.expense_line_ids.filtered("invoice_id")
+            if expenses_with_invoice:
+                if len(expenses_with_invoice) == len(sheet.expense_line_ids):
+                    sheets_to_skip_super += sheet
+                else:
+                    removed_expenses_map[sheet.id] = expenses_with_invoice
+                    sheet.expense_line_ids -= expenses_with_invoice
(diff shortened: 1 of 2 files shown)


dtec-landoo:
@pedrobaeza @victoralmau 

Hi! 

I've just updated the PR. All the CI checks, tests, and Codecov coverage are now completely green.

It is ready for review whenever you have a moment. Let me know if you need any further changes. Thanks!

dtec-landoo:
@pedrobaeza @victoralmau

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -19,37 +19,109 @@ def get_expense_sheets_with_invoices(self, func):
...
+                    )
+                    if ap_lines and transfer_line:
+                        (ap_lines + transfer_line).reconcile()
+
+    def _do_create_moves(self):
+        removed_expenses_map = {}
+        sheets_to_skip_super = self.env["hr.expense.sheet"]
+        all_generated_moves = self.env["account.move"]
+
+        # 1. Ocultar temporalmente los gastos con factura
```
pedrobaeza: Please don't put empty lines inside a method, and comment should be in English.

dtec-landoo:
@pedrobaeza @victoralmau 

Hi! It´s now correct? Thanks!

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -108,12 +152,10 @@ def _prepare_bills_vals(self):
             expenses_without_invoice = self.expense_line_ids.filtered(
                 lambda r: not r.invoice_id
             )
-            if expenses_without_invoice:
```
pedrobaeza, line 111: Why removing this if?

hr_expense_invoice/models/hr_expense_sheet.py
```diff
@@ -211,7 +253,7 @@ def _track_subtype(self, init_values):
             super()._track_subtype(init_values)
 
     def _check_can_create_move(self):
-        expense_sheets_with_invoices = self.get_expense_sheets_with_invoices(any)
+        expense_sheets_with_invoices = self.get_expense_sheets_with_invoices(all)
```
pedrobaeza, line 256: Why this change?

dtec-landoo:
@pedrobaeza  @victoralmau 

**Why removing this if?**

By removing this condition, we ensure that res['line_ids'] is explicitly overwritten, even if it results in an empty list. If a sheet contains only expenses with an associated invoice, keeping the if would leave res['line_ids'] with the values generated by super(). This would incorrectly generate standard journal entry lines for expenses that already have their own vendor bills. Removing the if forces it to be empty when needed, preventing duplicate accounting entries.

**Why this change?**

Changing any to all is necessary to correctly support mixed expense reports (containing both invoiced and non-invoiced expenses). If we use any, a mixed sheet is excluded from super()._check_can_create_move() and is treated as if it were fully invoiced, which leads to the 'Invalid Operation' error when trying to process the non-invoiced lines. By using all, only completely invoiced sheets pass the standard move creation checks, allowing mixed sheets to properly validate and generate the accounting entry for their non-invoiced lines.

pedrobaeza (approved):

victoralmau (approved):

PR #337, opened 2026-05-05
[19.0][WIP][ADD] hr_expense_trip: Link expenses with trips

CRogos:
Because hr.expense.sheet (Expense Reports) are discontinued in v19, this module adds a new possibility to link expenses to a trip.

@madmooose, @smaddlsoft your feedback is welcome


madmooose:
Here is my proposal
grafik

Not sure if i like the idea to skip the reports. For trips I would propose to get back the sheets logic and extend it with a trip module (date field and checker)

@smaddlsoft Trips was your initial requirement. What do you think?

hr_expense_trip/models/hr_expense.py
```diff
@@ -0,0 +1,110 @@
...
+            "employee_id": employee.id,
+            "start_date": min(expense_dates),
+            "end_date": max(expense_dates),
+            "state": "receipts",
+        }
+
+    def _prepare_receipts_vals(self):
+        return_vals = []
+        for (trip_id), expenses_sudo in (
+            self.sudo().grouped(lambda x: (x.trip_id)).items()
```
AnizR, line 66: I have some doubt about the sudo here.
Is it really necessary?

smaddlsoft:
@CRogos kindly rebase for testing, sorry for being late on this.

hr_expense_trip/models/hr_expense.py
CRogos: When you have a look in the hr_expense module in the same method, there is a similar situation:
https://github.com/odoo/odoo/blob/bce5c19eca5678e7f174d3c3785eb0712e36ccb4/addons/hr_expense/models/hr_expense.py#L1607

PR #341, opened 2026-05-09
[18.0][ADD] hr_expense_tax_distribution : add module

metaminux:
This module addresses a limitation of standard Odoo when dealing with mixed-rate receipts (e.g. restaurant bills in France, where food, soft drinks and alcoholic beverages are taxed at 5.5%, 10% and 20% respectively). In standard Odoo, an expense accepts a single set of taxes applied to the full amount, leading to an incorrect VAT breakdown in the accounting entry and wrong figures in the tax declaration.

When an expense carries more than one tax, a Tax Distribution table appears on the expense form. The user fills in the tax-excluded base amount for each applicable rate; tax amount and TTC total are computed automatically. The module validates that the distribution totals match the expense total before submission, and generates one base line and one tax line per distribution entry in the final accounting entry instead of a single blended line. Expenses with a single tax are unaffected.

### Key design decisions

- Distribution lines are managed via _sync_tax_distribution_lines(), a dedicated public method that downstream modules can override to add extra fields (e.g. a specific expense account per line) without touching the onchange.
- The submission constraint is enforced in action_submit_expenses rather than via @api.constrains, to avoid false positives during incremental onchange processing.


metaminux:
Hi @OCA/human-resources-maintainers 

Can someone review this PR please ?

Regards,

metaminux:
cc @sebastienbeau

metaminux:
Hi @sebalix and @florian-dacosta

Can you review this PR ?

metaminux:
Hi everyone,

Can someone give some approval to this PR ?

cc @sebastienbeau @sebalix @florian-dacosta @Kev-Roche

PR #345, opened 2026-05-24
[19.0][ADD] hr_expense_tax_distribution

dnplkndll:
Adds the new module `hr_expense_tax_distribution` on 19.0.

**Origin:** ported from an internal Akretion working branch authored by Guillaume Masson <guillaume.masson@akretion.com>. The module does not exist on any earlier OCA series (14.0–18.0), so this is a fresh `[ADD]` to OCA/hr-expense rather than a `[MIG]` from a prior series — the `[ADD]` prefix also routes `/ocabot merge` to the right (minor) bump.

## What it does

When a single expense receipt carries multiple VAT rates (a French restaurant bill at 5.5% / 10% / 20% is the canonical case), Odoo's stock `tax_ids` field can only carry one rate against the full total. This module adds a `tax_line_ids` one2many on `hr.expense` where the user splits the receipt total across as many tax-rate lines as needed, each holding a base amount and the applicable taxes; the resulting accounting move is built from those lines instead of the single `tax_ids` / total pair, producing a correct VAT breakdown. A `_constraint_total_matches` constraint blocks submit unless the distribution lines sum back to the expense total.

When no distribution lines are present (`has_tax_distribution = False`), the form hides the section and the module is a no-op — the expense flows through stock 19.0 paths unchanged.

- **Sheetless tax distribution.** The original Akretion design aggregated tax lines on `hr.expense.sheet`. Odoo 19 collapsed expense reports into expense lines, so this port overrides `_prepare_receipts_vals()` and `_prepare_payments_vals()` on `hr.expense` directly to inject the per-rate tax lines into the accounting entry without going through the (now-removed) sheet model.
- **Test modernisation.** Replaced legacy sheet workflows + action calls in the unit tests with direct line submission and direct posting (`_post_without_wizard`), and converted assertions to the 19.0 `expense.state` / `approval_state` machine.
- **Lazy translation refactor.** Standardised on `self.env._()` throughout instead of module-level `_()`.

## Test coverage

`tests/test_hr_expense_tax_distribution.py` — 22 test methods covering: compute paths on `hr.expense.tax.line` (single tax, multi-tax, zero base, no tax); the `has_tax_distribution` boolean; onchange rebuild semantics when `tax_ids` changes (adding, removing, preserving base amounts); the sum-matches constraint (matches, mismatches, zero-total, negative-base); and the accounting move generation (price_unit × quantity, account move amounts).


dnplkndll:
Replaced by #351 — source branch moved from dnplkndll fork to ledoent organization fork. Same commits (HEAD 4f05f21), no content change.

PR #351, opened 2026-05-25
[19.0][MIG] hr_expense_tax_distribution: Migration to 19.0

dnplkndll:
Port of `hr_expense_tax_distribution` to 19.0.

- **Sheetless tax distribution**: `hr.expense.sheet` is removed in 19.0, so the
  tax split is computed on `hr.expense` itself, overriding
  `_prepare_receipts_vals()` and `_prepare_payments_vals()`.
- **The two posting flows must not share a move-line builder.**
  `_prepare_receipts_vals` (employee-paid) posts an `in_receipt` — an invoice,
  where `price_unit`/`quantity`/`tax_ids` drive Odoo's invoice recompute.
  `_prepare_payments_vals` (company-paid) posts an `entry`, which takes
  `balance` verbatim and never expands `tax_ids` into tax lines. The
  company-paid flow therefore mirrors core's own tax computation, with one base
  line per distribution line rather than core's single one, and absorbs the
  rounding residue into the last base line so the entry balances exactly.
- **View placement**: the distribution table is anchored after the outer
  `<group>`. A `<group>` renders as a two-column label/field grid, so a
  block-level `div` as a direct child leaves an unpaired cell and shifts every
  label/field pair below it.
- **Test modernisation**: legacy sheet workflows replaced with direct line
  submission and posting. Tests cover both flows, including the company-paid
  journal entry.
- **Standardisation**: translations use `self.env._()`; long lines wrapped for
  the OCA ruff checks.

Assisted-by: Claude Opus 5


BhaveshHeliconia (approved):
Code review LGTM!

haivanlive:
Journal entry error occurs when using the payment method to Company and the payment type to Cash. Could you please check and fix for this?

dnplkndll:
> Journal entry error occurs when using the payment method to Company and the payment type to Cash. Could you please check and fix for this?

I think I got this fixed if you want to verify. thanks.

PR #352, opened 2026-05-25
[19.0][ADD] hr_expense_trip

dnplkndll:
Adds the new module `hr_expense_trip` on 19.0.

**Origin:** continuation of [#337](https://github.com/OCA/hr-expense/pull/337) by @CRogos, which marked the work as `[19.0][WIP][ADD]` and stopped short of a working 19.0 implementation. CRogos is credited via the `Co-authored-by:` trailer on the squash commit and listed in `readme/CONTRIBUTORS.md`. If/when this lands, #337 can be closed as superseded. The module does not exist on any earlier OCA series (verified 14.0–18.0 absent), so the prefix is `[ADD]` and not `[MIG]`.

## What it does

A trip is a date-bounded container that groups any number of `hr.expense` records belonging to the same employee. The trip carries a name, reason, partner, employee, start/end date, and a state machine (Draft → Requested → Collect Receipts → Done) with optional auto-approval via the `hr_expense_trip.auto_approve` system parameter. When the trip transitions to **Done**, a per-trip PDF report is rendered and posted to the trip's chatter as an attachment.

On the expense form, a `Trip` field lets the user link an expense to a matching trip; `default_get` pre-selects the matching open trip when the expense date falls inside it.

- **Sheetless container model.** Odoo 19 removed `hr.expense.sheet`. The trip plays the same "container for many expenses" role but with an independent approval state and an independent PDF report. The link is `hr.expense.trip_id` Many2one with `ondelete="set null"` so detaching the trip doesn't cascade-delete expenses.
- **PDF render uses the 19.0 `ir.actions.report._render_qweb_pdf(report_ref, res_ids)` signature** — the 18.0 record-bound `report._render_qweb_pdf(res_ids)` shape is gone. `_attach_trip_report` was updated accordingly.
- **`expense_lines_widget` ESM widget** keeps only the `isMany2Many=true` override so the user can add existing expenses via the standard m2m selector. The 18.0-era attachment-thread switching and FileViewer integration were tied to `hr.expense.sheet` (now removed) and have been dropped.
- **`static/src/views/list.xml`** (a verbatim copy of 18.0 core's sheet-mode list template) was deleted — the module's manifest asset glob is `*.js` only, so the file was never loaded, and every method it referenced (`displayCreateReport`, `displaySubmit`, …) is dead on 19.0.
- **Test gates the end-to-end state-machine path** including `action_done → _attach_trip_report → PDF attach via message_post`. The previous `patch.object(_attach_trip_report)` mock was removed — that mock was hiding the broken render signature from CI.

## Test coverage

`tests/test_hr_trip.py` — 12 test methods covering: trip creation defaults, employee default, date-range constraint (valid + invalid + start-after-end), `trip_id` cleared on parent unlink, "My Trips" search-domain filter, mail.thread mixin, full state-machine transitions (with real PDF render), `auto_approve` enabled vs disabled, and `default_get` trip pre-selection on the expense form (matching + non-matching cases).

### DIFF
--- a/hr_expense_trip/README.rst
+++ b/hr_expense_trip/README.rst
@@ -1,0 +1,68 @@
+==============
+HR Expense Trip
+==============
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
+    :target: https://odoo-community.org/page/development-status
+    :alt: Beta
+.. |badge2| image:: https://img.shields.io/badge/licence-AGPL--3-blue.png
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+.. |badge3| image:: https://img.shields.io/badge/github-OCA%2Fhr--expense-lightgray.png?logo=github
+    :target: https://github.com/OCA/hr-expense/tree/19.0/hr_expense_trip
+    :alt: OCA/hr-expense
+
+|badge1| |badge2| |badge3|
+
+This module allows you to group expenses under business trips.
+You can create a trip with a start date and end date, then attach expenses
+to it. The expense date column in the trip form is colour-coded:
+
+- **Green** – expense date is *before* the trip start date.
+- **Yellow** – expense date is *after* the trip end date.
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+#. Go to *Expenses → My Trips* to create and manage your own trips.
+#. Go to *Expenses → All Trips* (expense managers only) to see all trips.
+#. Open a trip and link expenses to it, or set the *Trip* field directly
+   on an expense form.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/hr-expense/issues>`_.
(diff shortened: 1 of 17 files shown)


CRogos:
@dnplkndll I am still continuing on the other PR. Feel free to make suggestions or pull requests when you have improvements. I "paused" to to collect feedback and because of other work, but I will continue soon. 

I think most (all) of your changes are by AI? Also not clear to me why to delete the readme folder?

dnplkndll:
@CRogos closing this. not sure why this got posted is just the same work. I was interested in using this as a dependency. In the past I was involved in some back card transaction sync modules and have just submitted the ramp and mercury api implementation, https://github.com/OCA/bank-statement-import/pull/950 https://github.com/OCA/bank-statement-import/pull/952 

once you associate the transactions to employees with the last 4, then you would want to assosiate to a PO or maybe an expense report trip so that all the trip tags might apply. 

i will have to workshop this again now. 

I can see why the change in core is it was clumsy I though for end users. and perhaps some agent wil take care of it in the future. but for now trips to budgest and approvals still seems like the right path.

PR #355, opened 2026-05-27
[19.0][MIG] hr_expense_payment: Migration to 19.0

dnplkndll:
Port of `hr_expense_payment` from 18.0 to 19.0 ([migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0)).

- Nothing stored anymore: `hr.expense.payment_ids` delegates to core `account_move_id.reconciled_payment_ids`; `account.payment.reconciled_expense_ids = (invoice_ids | reconciled_bill_ids).expense_ids` (core's `expense_ids` covers company-paid only). Register-wizard override and `post_init_hook` dropped; both fields `compute_sudo` + searchable.
- `migrations/19.0.1.0.0/post-migration.py` carries 18.0 `payment_expense_sheet_rel` rows whose reconciliation was since undone into core's `account_move__account_payment` via `former_sheet_id`; idempotent, no-op on fresh installs or when the pointer is unfilled.


BhaveshHeliconia (approved):
LGTM!

BhaveshHeliconia:
Please squash the commits as explained here: https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

hr_expense_payment/models/account_payment.py
```diff
@@ -8,12 +8,12 @@
 class AccountPayment(models.Model):
     _inherit = "account.payment"
 
-    expense_sheet_ids = fields.Many2many(
-        comodel_name="hr.expense.sheet",
-        relation="payment_expense_sheet_rel",
+    expense_ids = fields.Many2many(
```
pilarvargas-tecnativa: You are redefining this existing field in the model
https://github.com/odoo/odoo/blob/c47af9c0596a39a7a5d0d7b6709d2d796ab3d56e/addons/hr_expense/models/account_payment.py#L10

With these changes, the standard field is replaced by an independent Many2many using a new relation table. This duplicates the standard relationship and could leave both relations out of sync.

There is also no migration script for the existing data in payment_expense_sheet_rel.
A post_init_hook will only run when installing the module, not when upgrading an existing database, and the current implementation may also miss partially reconciled payments.

pilarvargas-tecnativa (changes requested):
@Tecnativa TT58634

Have you checked whether this functionality has already been incorporated into Odoo 19 core?

@victoralmau @carlos-lopez-tecnativa please, take a look

hr_expense_payment/models/account_payment.py
```diff
@@ -2,18 +2,37 @@
...
+    # company-paid expenses; this covers the reimbursement direction, derived
+    # from reconciliation like core's reconciled_bill_ids.
+    reconciled_expense_ids = fields.Many2many(
+        comodel_name="hr.expense",
+        string="Reimbursed Expenses",
+        compute="_compute_reconciled_expense_ids",
+        compute_sudo=True,
+        search="_search_reconciled_expense_ids",
+        help="Employee-paid expenses whose journal items have been reconciled "
+        "with this payment.",
```
pilarvargas-tecnativa: Thanks for the changes. However, please make sure that the functionality provided by this module has not already been incorporated into Odoo 19 core.

Also, the existing relationships stored in payment_expense_sheet_rel are not migrated. Since the new computed field is not stored, it only derives relationships from the available accounting data. Could you confirm that this reliably preserves all existing links, including historical ones?

The OpenUpgrade migration scripts for hr_expense may be useful as a reference for reviewing how the standard expense-payment relationships and their existing data are handled.

dnplkndll:
@pilarvargas-tecnativa 
hr_expense_cancel was adapted and I found no other consumers. So unless there is is a any other reason this would be useful to a UI or report it seems like we could drop it?

PR #356, opened 2026-05-27
[19.0][MIG] hr_expense_cancel: Migration to 19.0

dnplkndll:
Port of `hr_expense_cancel` from 18.0 to 19.0 ([migration guide](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0)).

- **Hook target changed** from `hr.expense.sheet.action_cancel` to `hr.expense.action_reset()` — core's [`action_reset`](https://github.com/odoo/odoo/blob/19.0/addons/hr_expense/models/hr_expense.py#L1210) refuses to reset an expense linked to a posted entry; the module tears down payments and the posted move first so the guard passes. Everything runs with the caller's rights, as in 18.0: payment teardown requires accounting ACL, and expenses without a move (the everyday employee reset) touch no accounting model.
- **No dependency on `hr_expense_payment`**: payments are found via core's `reconciled_payment_ids` (reconciliation walk ∪ stored `matched_payment_ids`, so wizard payments whose reconciliation was removed are torn down too), and `_remove_move_reconcile` walks partials via `_get_reconciled_amls()` instead of `full_reconcile_id`.
- **Per-expense, not per-sheet**: helpers operate on `expense.account_move_id`.
- **`payment.state` rename**: 18.0 `'cancel'` → 19.0 `'canceled'`.

## 19.0 relevance
Core's `action_reset` reverses the expense's move but refuses when payments exist; payment-side teardown is this module's remaining value.


pilarvargas-tecnativa (changes requested):
This migration still depends on hr_expense_payment and uses its custom hr.expense.payment_ids relation.

dnplkndll:
@pilarvargas-tecnativa updated to remove the hr_expense_payment dap and I think hr_expense_payment can be dropped

PR #358, opened 2026-05-28
[19.0][MIG] hr_expense_advance_clearing: Migration to 19.0

dnplkndll:
Port of `hr_expense_advance_clearing` from 18.0 to 19.0.

19.0 removed `hr.expense.sheet`, so advance and clearing move onto the expense itself: `expense_type` replaces the sheet-level `advance` flag, `clearing_advance_id` / `clearing_expense_ids` replace `advance_sheet_id` / `clearing_sheet_ids`, and clearing posts as a journal entry (Dr expense / Cr employee advance, capped at the advance's residual) reconciled against the advance rather than as a vendor bill.

`account.payment.register.advance_id` is a field rather than a context read: core calls `_init_payments()` under `clean_context()`, which strips every `default_*` key before the override runs.

Two defects that retarget introduced, both found by driving the flow in the UI rather than by the test suite:

- **Return Advance could not complete.** The wizard opened with no journal, so `currency_id` — computed from it — stayed empty and `account.move` rejected the payment. `default_get` now seeds the journal when `hr_return_advance` is set. The ORM path hid this because `currency_id` is `precompute=True`, so `create()` resolved a chain the dialog never could; the regression test therefore drives the wizard through `Form`.
- **The module's fields rendered without labels.** 18.0 placed them in a dedicated `<group>`; anchoring on `product_id position="after"` lands them inside core's anonymous `<div>`, where Odoo renders no label. Re-anchored as group children, with a test asserting placement.

The `[IMP]` commit adds demo data only: the module deliberately leaves the Employee Advance product's expense account unset (the right account is chart- and company-specific), which means on a fresh demo database — how reviewers meet it on runboat — the first advance fails `_check_advance` with "Employee advance product has no payable account" and the flow cannot be tried at all. The demo record creates a reconcilable current-asset account and wires it up. It never loads in production, and the suite passes identically with and without demo.

Replayed history carries the 18.0 fixes under their original authors, including #376 (draft clearing payments).


haivanlive:
Hi all,

Thank you for your hard work on migrating this module to 19.0! 

However, looking at the current implementation, forcing the clearing process entirely through the **Vendor Bill** workflow introduces several major operational limitations:

1. **Intermediary Receivable Accounts:** When an intermediary receivable account is involved in the advance-clearing flow, forcing a Vendor Bill workflow creates unnecessary friction and breaks the natural accounting logic.
2. **Journal Entries on Clearing:** When posting the clearing advance entry, it is automatically converted into a Vendor Bill. This results in an excessively high volume of journal entries and overly complex booking lines, which heavily complicates financial auditing and reconciliation.

Given that advance clearing should ideally be handled via flexible journal entries (`entry` type) rather than strictly as supplier invoices, could we please reconsider this design? Forcing everything into Vendor Bills makes the accounting distribution far more complex than it needs to be.

Looking forward to your thoughts on this!

dnplkndll:
@haivanlive I have added an adaptation https://github.com/OCA/hr-expense/pull/358/changes/ac285786ed3159fa541170cf830c9afc1fb908cb to try and accomplish this. let me know what you think

haivanlive:
Hi @dnplkndll,

Thank you for your hard work on migrating this module to 19.0! Here are some feedback and issues I encountered during testing:
1. Clear Advance
Status: Working as expected. The clearing process via journal entries looks good.

2. Employee Advance Journal & Purchase Receipt Issue
Issue: When creating an Employee Advance and posting the journal entry, Odoo forces a wizard selecting the Purchase journal and generates a Purchase Receipt (in_receipt).
Accounting Impact: This is accounting-wise incorrect. Employee advances should not be treated as a commercial supplier invoice/purchase receipt.
Suggestion: We should have an option (or default behavior) to route the Employee Advance creation into a **Miscellaneous Operations journal** (General Journal) instead of forcing a Purchase journal. 
Proposed Entry: The posted journal entry should be a clean miscellaneous entry:
- Debit: Employee Advance account
- Credit: Outstanding Payments
Once this entry is fully reconciled/matched with the bank payment lines, the status of the Employee Advance record should automatically transition to Paid.

3. Return Advance Bug
Error: When trying to process a Return Advance, the system throws the following traceback:

"The register payment wizard should only be called on account.move or account.move.line records."

Saran440:
@dnplkndll Hi, thank you for you migrate to version19.
In v18 has fixed bug 3 issue, should you add it in your commit please?
- https://github.com/OCA/hr-expense/pull/370
- https://github.com/OCA/hr-expense/pull/368
- https://github.com/OCA/hr-expense/pull/369

imbarbudiman:
@dnplkndll Field account.payment.advance_id is never populated, because _init_payments() call with [clean_context()](https://github.com/odoo/odoo/blob/b369bef588a09e631270a0b55970d4268f309728/addons/account/wizard/account_payment_register.py#L1296)

image

Saran440:
@dnplkndll @haivanlive @imbarbudiman 

Hi,
I think we should discuss the functional concept of advance/clearing before merging this migration.

In my opinion, one advance request should be able to cover multiple purposes.

For example, an employee requests an advance of 10,000:
- 3,000 for general expenses
- 5,000 for hotel
- 2,000 for fuel

In this case, I would expect the advance to contain 3 lines. When clearing the advance, each clearing expense should be linked to the corresponding advance line.

Since `hr.expense.sheet` was removed in Odoo 19, the current migration changes the concept to one advance per `hr.expense`.

I see 3 possible approaches:
1. Create a new model specifically for Advance / Clearing, which acts as the parent document and contains multiple advance lines.
2. Depend on #378 (hr_expense_sheet) and keep a concept closer to the previous versions.
3. Keep `hr_expense_advance_clearing` independent and add a separate integration module, e.g. `hr_expense_advance_clearing_sheet`, depending on both `hr_expense_advance_clearing` and `hr_expense_sheet`.


What do you think about this approach?

imbarbudiman:
> I see 3 possible approaches:
> 
> 1. Create a new model specifically for Advance / Clearing, which acts as the parent document and contains multiple advance lines.
> 2. Depend on [[19.0][ADD] hr_expense_sheet: New Module #378](https://github.com/OCA/hr-expense/pull/378) (hr_expense_sheet) and keep a concept closer to the previous versions.
> 3. Keep `hr_expense_advance_clearing` independent and add a separate integration module, e.g. `hr_expense_advance_clearing_sheet`, depending on both `hr_expense_advance_clearing` and `hr_expense_sheet`.
> 
> What do you think about this approach?

I'm currently working on a 16 → 19 upgrade. On 16 we had already customized the Advance vs Clearing relation into an m2m (originally o2m). 
So when something this big changes in core, like dropping hr.expense.sheet, it ends up requiring a lot of rework, because Advance / Clearing depends heavily on hr.expense.sheet.

So, regardless of the new hr_expense_sheet module, the idea of ​​creating a dedicated new model for Advance / Clearing is a good one. This way, if there are major changes to the Odoo core in the future, Advance / Clearing won't be affected as much.

dnplkndll:
I saw the sheet re-introduction. 
I would be for 
3) Keep hr_expense_advance_clearing independent and add a separate integration module, e.g. hr_expense_advance_clearing_sheet, depending on both hr_expense_advance_clearing and hr_expense_sheet

Saran440:
@dnplkndll Thanks for your works. I will review this module in the next week.

Saran440:
> @dnplkndll Field account.payment.advance_id is never populated, because _init_payments() call with [clean_context()](https://github.com/odoo/odoo/blob/b369bef588a09e631270a0b55970d4268f309728/addons/account/wizard/account_payment_register.py#L1296)
> 
> image

I'm not sure this review is fixed?

dnplkndll:
> I'm not sure this review is fixed?

ah right should be now. sorry for the chatter my picks were out of order. 

the return advance perms under Full Accounting Features is the design I guess?

Saran440:
@dnplkndll  Functional Review:
1. Add new menu My Advance. When select this menu, It should default type following menu but allow change.
Screenshot 2026-08-24 091101

2. Change type on form to radio Horizontal, And delete price duplicate
Screenshot 2026-08-24 094255

3. Return Advance should be change amount.
Example: Advance 1000, We need return 200. I can't change amount. and when create payment it error.
Screenshot 2026-08-24 094714
Screenshot 2026-08-24 094750

4. Button `clear advance` on advance document is disappear
5. Below Category on type expense is advance clearing? Should we move it to right position?
Screenshot 2026-08-24 095721


Since there are quite a few major changes in this version, would it be okay if I take care of this one?

dnplkndll:
> Since there are quite a few major changes in this version, would it be okay if I take care of this one?

sure thanks!

imbarbudiman:
@Saran440 
@dnplkndll 
Wouldn't it be more appropriate for `clearing_residual` value to 0 before posting?

clearing_residual before posting

Saran440:
@imbarbudiman Yes, It should have value after register advance. I still working on this module.

Saran440:
@imbarbudiman @dnplkndll 

I've created a new PR for this module in https://github.com/OCA/hr-expense/pull/358. Could you please close this PR and continue testing/reviewing on https://github.com/OCA/hr-expense/pull/358 instead?

imbarbudiman:
> I've created a new PR for this module in #358. Could you please close this PR and continue testing/reviewing on #358 instead?

You mean #380, right?

Saran440:
Yes, #380 . Sorry, my mistake.

PR #353, opened 2026-05-26
[19.0][MIG] hr_expense_sequence: Migration to 19.0

dnplkndll:
Port of `hr_expense_sequence` from 18.0 to 19.0 ([migration guide](https://github.com/OCA/maintainer-tools/wiki/#migration-to-version-19.0)).

## Non-mechanical adaptations worth flagging

- **Sheet removed in 19.0**: `_inherit = "hr.expense.sheet"` → `_inherit = "hr.expense"`. The `number` Char + auto-sequence-in-`create()` logic moves to `hr.expense` itself, and `_rec_name = "number"` follows.
- **The sequence record keeps its 18.0 code and name** (`hr.expense.sheet` / *Expense report sequence*), even though the string names a model that no longer exists. The data file is `noupdate="1"`, so a database upgraded from 18.0 keeps the record as created; renaming it would leave those databases without a sequence to draw from and every expense would keep the `/` placeholder. A test pins the code to the shipped record. Happy to swap this for a rename plus a migration script if maintainers prefer that direction.
- **View inherits**: rewired to `hr_expense.hr_expense_view_form` / `hr_expense_view_expenses_analysis_tree` / `hr_expense_view_search` (the 18.0 sheet view IDs are gone). `number` is on both the list and the form.
- **Report inherit**: rewired to `hr_expense.report_expense` (was `report_expense_sheet`).
- **`post_init_hook`**: backfill now iterates `hr.expense` records (not sheets); same sequence-pull pattern.
- **Test rewrite**: 18.0 tests created sheets and asserted the number on the sheet; the 19.0 version exercises `hr.expense.create()` with explicit / auto / manual-override paths.
- **Translations are carried verbatim from 18.0** — no `.pot`/`.po` regeneration in this PR, so Weblate contributions (including the Slovenian and Turkish files) survive the port untouched.

AI-assisted (Claude Code); every change reviewed, tested, and owned by the author.

PR #359, opened 2026-05-28
[19.0][MIG] hr_expense_advance_clearing_sequence: Migration to 19.0

dnplkndll:
Port of `hr_expense_advance_clearing_sequence` from 18.0 to 19.0. Gives advance expenses their own `AV` numbering, separate from the generic expense sequence.

Depends on [OCA/hr-expense#353](https://github.com/OCA/hr-expense/pull/353) (`hr_expense_sequence`) + [OCA/hr-expense#358](https://github.com/OCA/hr-expense/pull/358) (`hr_expense_advance_clearing`). Both pinned in `test-requirements.txt`, which is what reds `Detect unreleased dependencies`.

- **Inherit `hr.expense`** (was `hr.expense.sheet`); key off `expense_type == 'advance'` (was the sheet's `advance` Boolean), and honour a `default_expense_type` context default the way 18.0 honoured `default_advance`.
- **Sequence code and name deliberately keep their 18.0 values** (`hr.expense.sheet.advance` / *Expense report sequence (Advance)*), even though the string names a model that no longer exists. The data file is `noupdate="1"`, so a database upgraded from 18.0 keeps the record as created; renaming it here would leave those databases with a sequence `create()` can no longer find, and advances would silently fall back to the generic expense number. A test pins the code to the shipped record.
- Test subclasses `hr_expense_advance_clearing`'s `TestHrExpenseAdvanceClearing`, inheriting its fixtures and adding the numbering cases (advance gets an `AV` number, context-defaulted advance gets one too, regular expense does not).


BhaveshHeliconia (approved):
Code review LGTM!

BhaveshHeliconia (approved):
LGTM!

PR #360, opened 2026-06-03
[18.0][IMP] hr_expense_tier_validation: clear review_ids when reset draft & exception expense

Saran440:
This PR fixed & implement in `hr_expense_tier_validation`

### 1. Approve expense sheet and Reset to Draft

**Step to reproduce:**

1. Create tier validation with hr.expense.sheet
2. Create expense and approve (with Tier)
3. Reset to draft expense, state change to To Submit but user can't editable name
    
selection_290

### 2. Tier Validation Exception on `hr.expense`

This case improved to use standard, because new version base_tier_validation has `Tier Validation Exception`. So, we change exception from hard code to config following standard

**Changes:**
- Add exception config on `hr.expense`
- After validated, `hr.expense` is read-only unless exception is configured

selection_291

### DIFF
--- a/hr_expense_tier_validation/README.rst
+++ b/hr_expense_tier_validation/README.rst
@@ -42,14 +42,26 @@ Configuration
 To configure tier validation for HR expenses:
 
 1. Go to *Settings > Technical > Tier Validations > Tier Definition*
-2. Create the desired approval tiers for the hr.expense model.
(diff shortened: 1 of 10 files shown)


ROBBYHOOD9 (approved):
LGTM 👍

Pani-k-folk (approved):
Code LGMT 👍

PR #368, opened 2026-06-22
[18.0][FIX] hr_expense_advance_clearing: add analytic line advance when clearing

Saran440:
This PR adds analytic distribution to the journal entry lines created during the clearing process.

Step to reproduce:
1. Create advance with analytic
2. Create clearing with analytic
3. The journal entry lines generated from the advance in the clearing process have an empty analytic distribution.

The journal entry lines should inherit the analytic distribution from the original advance.
Screenshot 2026-06-22 133940


ROBBYHOOD9 (approved):
LGTM 👍

Pani-k-folk (approved):
Code LGTM 👍

PR #370, opened 2026-06-23
[18.0][FIX] hr_expense_advance_clearing: permission return_count

Saran440:
This PR fixed users expense can't submit expense when install `hr_expense_advance_clearing` because field `return_count` count from `account.payment`

Step to reproduce:
1. install module hr_expense_advance_clearing
2. login with user without accounting
3. create expense > Submit > it will error permission
selection_429


ROBBYHOOD9 (approved):
LGTM 👍

Pani-k-folk (approved):
LGTM 👍

PR #371, opened 2026-06-23
[18.0][IMP] hr_expense_petty_cash: fix journal items on petty cash with included tax and analytic

Saran440:
This PR fixed 2 case
1. Send analytic distribution to journal entry of petty cash
2. Fixed line included tax following expense standard

Before
selection_432

After
Screenshot 2026-06-23 200155

### DIFF
--- a/hr_expense_petty_cash/models/account_move.py
+++ b/hr_expense_petty_cash/models/account_move.py
@@ -15,6 +15,12 @@ class AccountMove(models.Model):
         self._check_petty_cash_amount()
         return super().action_post()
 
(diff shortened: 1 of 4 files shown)


Saran440:
@BhaveshHeliconia Hi! Since you migrated this module to v18, would you be okay if we changed these lines?

I think this is the standard approach. What do you think?

BhaveshHeliconia (approved):
I tested this PR on an Odoo 18.0 database, and it works as expected. The refactored code uses Odoo 18's native journal entry engine to handle the tax split correctly.

ROBBYHOOD9 (approved):
LGTM

PR #372, opened 2026-06-24
[18.0][ADD] hr_expense_duplicate_disable

Saran440:
This module add config to disable duplicate wizard when approved document expense

Use concept from https://github.com/OCA/hr-expense/pull/173 and rename

selection_437

Before config, It will display new wizard when duplicate document
Screenshot 2026-06-24 175816


Pani-k-folk (approved):
CODE LGTM 👍

PR #378, opened 2026-08-13
[19.0][ADD] hr_expense_sheet: New module

victoralmau:
New Module

Related to https://github.com/odoo/odoo/pull/189701

Please @carlos-lopez-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT58631


Saran440:
@victoralmau Thanks for your work.
When I **add a line** to the expense report, the `Upload` and `New` buttons should not appear at the top.
Screenshot 2026-08-18 085709

pedrobaeza:
Trying to load demo data, I found this problem:

imagen

It's also curious that the runboat didn't contain any installed module. That happens AFAIK when some module can't be loaded.

victoralmau:
> Trying to load demo data, I found this problem:

Fixed

haivanlive:
I cannot find the 'New' button in the list view to create a new record hr.expense.sheet. Please check and fix this issue when you have a moment.

image

victoralmau:
> I cannot find the 'New' button in the list view to create a new record hr.expense.sheet. Please check and fix this issue when you have a moment.

Fixed!

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,777 @@
...
+                    <button
+                        name="action_approve"
+                        string="Approve"
+                        type="object"
+                        data-hotkey="q"
+                        context="{'validate_analytic': True}"
+                        invisible="not can_approve or state != 'submitted'"
+                        class="oe_highlight o_expense_sheet_approve"
+                    />
+                    <button
```
carlos-lopez-tecnativa, line 92: What happened to the Pay button?
https://github.com/odoo/odoo/blob/83c8af1fdd026d73f70adc6531d79de1e2e0803f/addons/hr_expense/views/hr_expense_views.xml#L772

Image

hr_expense_sheet/views/hr_expense_views.xml
```diff
@@ -0,0 +1,66 @@
...
+            </xpath>
+            <xpath expr="(//button[@name='action_post'])[2]" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </xpath>
+            <button name="action_refuse" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </button>
+            <field name="manager_id" position="attributes">
+                <attribute name="invisible">1</attribute>
+            </field>
```
carlos-lopez-tecnativa, line 54: I think these buttons should not be hidden only in the specific one2many expense.sheet view, but also in the general expense view. When I go to the expense form view, I can still see the buttons.

I have in mind that if an expense doesn't have a sheet, the expense follows its own flow. However, if the expense is part of a sheet, then the flow is handled by the sheet. With this in mind, the buttons should be hidden when a sheet exists. What do you think?

Image

hr_expense_sheet/models/hr_expense.py
```diff
@@ -0,0 +1,300 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+from odoo import Command, api, fields, models
+from odoo.exceptions import UserError
+from odoo.tools import float_round
+
+
+class HrExpense(models.Model):
+    _inherit = "hr.expense"
```
carlos-lopez-tecnativa, line 9: Currently, I don't have a way to differentiate whether my expense is part of a sheet or not.

In V18, we had states. I understand that adding new states to the flow could be complicated, but at least we could add some visual indicators in the list view (add the `shet_id` as optional field) and filters in the search view to make it easy to differentiate between them.

For example:
https://github.com/odoo/odoo/blob/83c8af1fdd026d73f70adc6531d79de1e2e0803f/addons/hr_expense/views/hr_expense_views.xml#L402
Image

hr_expense_sheet/views/hr_expense_views.xml
```diff
@@ -0,0 +1,66 @@
...
+            </xpath>
+            <xpath expr="(//button[@name='action_submit'])[1]" position="before">
+                <button
+                    name="action_submit_expenses"
+                    string="Create Report"
+                    type="object"
+                    class="oe_highlight o_expense_submit"
+                    invisible="nb_attachment &lt;= 0 or sheet_id"
+                    data-hotkey="v"
+                />
```
carlos-lopez-tecnativa, line 18: When I don't have an attachment, I can't create the report from the form view. In V18, this is possible because Odoo has two buttons. Please consider adding the button:
https://github.com/odoo/odoo/blob/83c8af1fdd026d73f70adc6531d79de1e2e0803f/addons/hr_expense/views/hr_expense_views.xml#L135
Image

carlos-lopez-tecnativa (changes requested):

hr_expense_sheet/views/hr_expense_sheet_views.xml
victoralmau: It doesn't exist because it doesn't exist in the expenses in v19 either; the same applies to the sheet status—they are consistent and match the expenses.

carlos-lopez-tecnativa: Okay, thanks. Now I see that the expenses are marked as paid when the bill is paid.

hr_expense_sheet/views/hr_expense_views.xml
```diff
@@ -0,0 +1,66 @@
...
+                <button
+                    name="action_view_sheet"
+                    type="object"
+                    string="View Report"
+                    class="oe_highlight"
+                    invisible="not sheet_id or nb_attachment &lt; 1"
```
carlos-lopez-tecnativa, line 24: If I have a sheet_id but no attachment, I don't see this button. Why?

What is the logic here?

Is it necessary to always upload an attachment to use the sheet? If so, raise a UserError when the user tries to create a report without an attachment, or allow the user to see this button even without an attachment.

I created an expense and created the report from the list view, but now I don't know whether my expense has a sheet or not. Please add the button, the field, or both.

Image

carlos-lopez-tecnativa (changes requested):

hr_expense_sheet/models/hr_expense.py
victoralmau: I added the `sheet_id` field to the list view as optional, and the “To report” filter has been added.

hr_expense_sheet/views/hr_expense_views.xml
victoralmau: Sorry, it's already been changed. It's similar to v18 now, and you don't have to attach anything if you don't want to.

hr_expense_sheet/views/hr_expense_views.xml
victoralmau: Sorry, it's already been changed. It's similar to v18 now, and the “View report” button is always displayed.

hr_expense_sheet/views/hr_expense_views.xml
victoralmau: You're right—it's already been changed, and it's hidden if a `sheet_id` is set.

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,777 @@
...
+                <filter
+                    string="My Reports"
+                    name="my_reports"
+                    domain="[('employee_id.user_id', '=', uid)]"
+                />
+                <filter
+                    string="My Team"
+                    name="my_team_reports"
+                    domain="[('employee_id.parent_id.user_id', '=', uid)]"
+                    groups="hr_expense.group_hr_expense_manager"
```
carlos-lopez-tecnativa: A user with the `hr_expense.group_hr_expense_team_approver` group could see the expenses where they are the manager. So, it makes more sense to me for this filter to be restricted to this group. What do you think?

Image

```suggestion
                    groups="hr_expense.group_hr_expense_team_approver"
```

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,777 @@
...
+            </xpath>
+        </field>
+    </record>
+    <record id="view_hr_expense_sheet_pivot" model="ir.ui.view">
+        <field name="name">hr.expense.sheet.pivot</field>
+        <field name="model">hr.expense.sheet</field>
+        <field name="arch" type="xml">
+            <pivot string="Expenses Analysis" sample="1">
+                <field name="employee_id" type="row" />
+                <field name="accounting_date" interval="month" type="col" />
```
carlos-lopez-tecnativa, line 433: I got an error with this field. I'm not sure if it's caused by the demo data, but it seems that from the expense.sheet pivot view, you can navigate to expenses, and this field doesn't exist on that model.

Could you please review this? I've attached a video.

https://github.com/user-attachments/assets/45d79d8f-6fbf-48fe-823c-2ee1efaa8de0

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,777 @@
...
+    <record id="view_hr_expense_sheet_dashboard_tree_header" model="ir.ui.view">
+        <field name="name">hr.expense.sheet.dashboard.list</field>
+        <field name="model">hr.expense.sheet</field>
+        <field name="priority">20</field>
+        <field name="inherit_id" ref="view_hr_expense_sheet_tree" />
+        <field name="mode">primary</field>
+        <field name="arch" type="xml">
+            <xpath expr="//list" position="attributes">
+                <attribute name="js_class">hr_expense_dashboard_tree</attribute>
+            </xpath>
```
carlos-lopez-tecnativa: I'm not sure if the error comes from trying to use this JS class, which belongs to `hr.expense`, from `hr.expense.sheet`. However, when I click on any action in the dashboard, I get this error.

Image

hr_expense_sheet/views/hr_expense_sheet_views.xml
carlos-lopez-tecnativa: The same happens in the calendar view.

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,777 @@
...
+        <field name="domain">[]</field>
+        <field name="help" type="html">
+            <p class="o_view_nocontent_smiling_face">
+            Create a new expense report
+          </p>
+            <p>
+            Once you have created your expense, submitted it to your manager who will validate it.
+          </p>
+        </field>
+    </record>
```
carlos-lopez-tecnativa: Are these actions used?

hr_expense_sheet/models/account_move.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    @api.depends("commercial_partner_id")
+    def _compute_show_commercial_partner_warning(self):
+        for move in self:
+            move.show_commercial_partner_warning = (
+                move.commercial_partner_id == self.env.company.partner_id
+                and move.move_type == "in_invoice"
+                and move.partner_id.sudo().employee_ids
+            )
```
carlos-lopez-tecnativa, line 27: This field is used in expenses? The name looks too generic in this context.

Since this field is not stored, could it be renamed to something more specific to the expenses context?

I'm not sure whether this warning is useful.

carlos-lopez-tecnativa (changes requested):

hr_expense_sheet/views/hr_expense_sheet_views.xml
victoralmau: Fixed.

hr_expense_sheet/views/hr_expense_sheet_views.xml
victoralmau: Removed; they are no longer needed (although they existed in v18).

hr_expense_sheet/views/hr_expense_sheet_views.xml
victoralmau: Fixed. Actually, in v18, that data was displayed in `hr.expense` and `hr.expense.sheet`, but I think that's incorrect and also problematic now (it has been removed).

hr_expense_sheet/models/account_move.py
victoralmau: This is a feature that was available in v18 and displays a warning on the invoice when applicable: https://github.com/odoo/odoo/blob/c11ca6d10cac99b367078823dafef46808346267/addons/hr_expense/views/account_move_views.xml#L26

hr_expense_sheet/views/hr_expense_views.xml
```diff
@@ -0,0 +1,144 @@
...
+                    name="sheet"
+                    domain="[]"
+                    context="{'group_by': 'sheet_id'}"
+                />
+            </filter>
+        </field>
+    </record>
+    <record id="hr_expense.menu_hr_expense_expenses_to_process" model="ir.ui.menu">
+        <field name="action" eval="False" />
+    </record>
```
carlos-lopez-tecnativa, line 143: I can understand the reason for archiving this menu at the moment, but please add a comment to document the reason and help us remember it in the future.

However, this menu could be useful for expense sheets instead of expenses. What do you think?

Image

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,691 @@
...
+        name="My Reports"
+    />
+    <menuitem
+        id="menu_hr_expense_report"
+        name="Expense Reports"
+        sequence="2"
+        parent="hr_expense.menu_hr_expense_root"
+        action="action_hr_expense_sheet_all"
+        groups="account.group_account_user,hr_expense.group_hr_expense_team_approver"
+    />
```
carlos-lopez-tecnativa, line 690: I suggest putting this menu under the `Reporting` menu. What do you think? Or is there any reason to keep it here?

Image

```suggestion
    <menuitem
        id="menu_hr_expense_report"
        name="Expense Sheets to process"
        sequence="2"
        parent="hr_expense.menu_hr_expense_reports"
        action="action_hr_expense_sheet_all"
        groups="account.group_account_user,hr_expense.group_hr_expense_team_approver"
    />
```

carlos-lopez-tecnativa: I noticed that this menu action is filtered by `Draft` and `Submitted` by default, which looks similar to my suggestion in this comment: https://github.com/OCA/hr-expense/pull/378/changes#r3881933520

What do you think about putting this menu in place of `Expenses to Process`? That could be another option.

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,691 @@
...
+                <field string="Journal" name="journal_id" />
+                <filter
+                    string="My Reports"
+                    name="my_reports"
+                    domain="[('employee_id.user_id', '=', uid)]"
+                />
+                <filter
+                    string="My Team"
+                    name="my_team_reports"
+                    domain="[('employee_id.parent_id.user_id', '=', uid)]"
```
carlos-lopez-tecnativa: I noticed a distinction in Odoo between the Manager (`parent_id`) and the Expense Approver (`expense_manager_id`). However, if you review the code, expense_manager_id takes priority over parent_id for notifying activities and determining who can approve the expense.

So, this filter should use `expense_manager_id` instead of `parent_id`, or ideally, both fields. What do you think about this logic?

Here are some references where `expense_manager_id` takes priority over `parent_id`:

https://github.com/odoo/odoo/blob/827d2b0ae048f9f6f304d3c530cf73bdf8fb8a63/addons/hr_expense/models/hr_expense.py#L81-L85
https://github.com/odoo/odoo/blob/827d2b0ae048f9f6f304d3c530cf73bdf8fb8a63/addons/hr_expense/models/hr_expense.py#L362
https://github.com/odoo/odoo/blob/827d2b0ae048f9f6f304d3c530cf73bdf8fb8a63/addons/hr_expense/models/hr_expense.py#L791
https://github.com/odoo/odoo/blob/827d2b0ae048f9f6f304d3c530cf73bdf8fb8a63/addons/hr_expense/models/hr_expense.py#L1403
https://github.com/odoo/odoo/blob/827d2b0ae048f9f6f304d3c530cf73bdf8fb8a63/addons/hr_expense/models/hr_expense.py#L1514

hr_expense_sheet/security/ir_rule.xml
```diff
@@ -0,0 +1,55 @@
...
+<odoo noupdate="1">
+    <record id="ir_rule_hr_expense_sheet_manager" model="ir.rule">
+        <field name="name">Manager Expense Sheet</field>
+        <field name="model_id" ref="model_hr_expense_sheet" />
+        <field name="domain_force">[(1, '=', 1)]</field>
+        <field
+            name="groups"
+            eval="[
+            (4, ref('account.group_account_user')),
+            (4, ref('hr_expense.group_hr_expense_user'))]"
```
carlos-lopez-tecnativa: Could you change the old syntax to the new `Command` syntax in all places?
```suggestion
            Command.link(ref('account.group_account_user')),
            Command.link(ref('hr_expense.group_hr_expense_user'))]"
```

hr_expense_sheet/static/src/views/expense_line_widget.esm.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /** @override **/
+    async onGlobalClick(ev) {
+        const expenseName = this.props.record.data.name;
+        const attachments = await this.orm.call(
+            "hr.expense",
+            "get_expense_attachments",
+            [this.props.record.resId]
+        );
```
carlos-lopez-tecnativa: I get an error when I click on the Kanban view. Please review this.
Image

Image

hr_expense_sheet/__manifest__.py
carlos-lopez-tecnativa, line 1: Could you please add an icon for the module? The Activities systray would look better with one.

Image

carlos-lopez-tecnativa (changes requested):

hr_expense_sheet/views/hr_expense_views.xml
victoralmau: That menu doesn't exist in v18; it's in v19 to differentiate between the expenses to be submitted, but if you're using this module, it's precisely to perform that action from the sheet; that's why it's “_hidden_.”

In my opinion, displaying that menu (which didn't exist in v18) to do something “differently” (submitting expenses individually, NOT through sheets) would be completely wrong.

carlos-lopez-tecnativa: > That menu doesn't exist in v18; it's in v19 to differentiate between the expenses to be submitted, but if you're using this module, it's precisely to perform that action from the sheet; that's why it's “_hidden_.”
> 
> In my opinion, displaying that menu (which didn't exist in v18) to do something “differently” (submitting expenses individually, NOT through sheets) would be completely wrong.

Sorry, I didn't mean that this menu should display the expenses. I meant that this menu should display the **Expense Sheets to process**.

However, if you prefer to keep this menu archived, that's okay. In that case, please take into account my other suggestion: https://github.com/OCA/hr-expense/pull/378#discussion_r3881997808

hr_expense_sheet/__manifest__.py
victoralmau: Okay, I've changed it—I added the same icon as `hr_expense`.

hr_expense_sheet/hooks.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    env.cr.execute(
+        """
+        SELECT column_name
+        FROM information_schema.columns
+        WHERE table_name='hr_expense_sheet' AND
+        column_name='payment_state'
+        """
+    )
+    if env.cr.fetchone():
```
carlos-lopez-tecnativa: You can use [column_exists](https://github.com/OCA/openupgradelib/blob/a6134fa5d056f10f12aa6206a022dbdc5c0540b8/openupgradelib/openupgrade_tools.py#L38) from `openupgradelib` or [column_exists](https://github.com/odoo/odoo/blob/fcdd41ff24e645c353f7e5276aa98952f3208462/odoo/tools/sql.py#L338) from Odoo's core tools.

hr_expense_sheet/hooks.py
```diff
@@ -0,0 +1,115 @@
...
+    has_former_sheet = env.cr.fetchone()[0]
+    if not has_former_sheet:
+        return
+    # Change states
+    env.cr.execute(
+        "UPDATE hr_expense_sheet SET state = 'submitted' WHERE state = 'submit'"
+    )
+    env.cr.execute(
+        "UPDATE hr_expense_sheet SET state = 'approved' WHERE state = 'approve'"
+    )
```
carlos-lopez-tecnativa, line 25: I left this comment https://github.com/OCA/OpenUpgrade/pull/5917#discussion_r3830155270, and you made the changes, but here you reverted them. Why?

carlos-lopez-tecnativa (changes requested):
LGTM, just a minor comment to understand the reason.

hr_expense_sheet/hooks.py
victoralmau: Related to https://github.com/OCA/OpenUpgrade/pull/5917#pullrequestreview-5027489803

carlos-lopez-tecnativa: Hmm, I understand the reason for moving the migration script from OpenUpgrade to here, but my comment was about using some functions from openupgradelib. Whether the code is here or in OpenUpgrade is not a reason to avoid using utility functions that make the code more readable and prevent code duplication.

victoralmau: "_In the module, you should also include a migration script without @openupgrade.migrate decorator importing and calling pre_init_hook._" 

This is why I don't use functions from openupgradelib

carlos-lopez-tecnativa: But the [map_values](https://github.com/OCA/openupgradelib/blob/a6134fa5d056f10f12aa6206a022dbdc5c0540b8/openupgradelib/openupgrade.py#L2259) function doesn't need an `env` variable; it only needs the cursor (`cr`). So, I think it is possible to call the function even though the method is not decorated with `@openupgrade.migrate`. What do you think, @pedrobaeza?

hr_expense_sheet/static/src/views/list.esm.js
```diff
@@ -0,0 +1,109 @@
...
+import {user} from "@web/core/user";
+
+patch(ExpenseListController.prototype, {
+    displayCreateReport() {
+        const usesSampleData = this.model.useSampleModel;
+        const records = this.model.root.records;
+        return (
+            !usesSampleData &&
+            records.length &&
+            records.some((record) => record.data.state === "draft")
```
carlos-lopez-tecnativa: Could you filter this to only include records where `sheet_id` is not set?

Image

hr_expense_sheet/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,798 @@
...
+            sheet.accounting_date = sheet._geet_default_accounting_date()
+        sheets._do_create_moves()
+
+    def action_reset(self):
+        self.expense_line_ids.action_reset()
+        self._do_reset_approval()
+
+    def action_open_expense_view(self):
+        self.ensure_one()
+        return self.expense_line_ids._get_records_action()
```
carlos-lopez-tecnativa: Please add a `name` to the action to prevent it from being displayed as `Unnamed`.
Image


```suggestion
        return self.expense_line_ids._get_records_action(name=self.env_("Expenses"))
```

hr_expense_sheet/views/hr_expense_sheet_views.xml
```diff
@@ -0,0 +1,691 @@
...
+                <field string="Journal" name="journal_id" />
+                <filter
+                    string="My Reports"
+                    name="my_reports"
+                    domain="[('employee_id.user_id', '=', uid)]"
+                />
+                <filter
+                    string="My Team"
+                    name="my_team_reports"
+                    domain="[('employee_id.expense_manager_id', '=', uid)]"
```
carlos-lopez-tecnativa, line 468: We changed the domain in the `My Team` filter to use `expense_manager_id`, as I suggested in https://github.com/OCA/hr-expense/pull/378#discussion_r3882072788. I think we should update the filter in Expenses to be consistent on both sides.

Alternatively, this could be a PR to Odoo. What do you think?

https://github.com/odoo/odoo/blob/16aaaafd7d314c04f39b1f633af3b5e15b46d78b/addons/hr_expense/views/hr_expense_views.xml#L423

carlos-lopez-tecnativa (changes requested):

hr_expense_sheet/views/hr_expense_sheet_views.xml
victoralmau: https://github.com/odoo/odoo/pull/287768

hr_expense_sheet/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,798 @@
...
+            sheet.accounting_date = sheet._geet_default_accounting_date()
+        sheets._do_create_moves()
+
+    def action_reset(self):
+        self.expense_line_ids.action_reset()
+        self._do_reset_approval()
+
+    def action_open_expense_view(self):
+        self.ensure_one()
+        return self.expense_line_ids._get_records_action(name=self.env_("Expenses"))
```
carlos-lopez-tecnativa:
```suggestion
        return self.expense_line_ids._get_records_action(name=self.env._("Expenses"))
```

Image

hr_expense_sheet/static/src/views/list.esm.js
```diff
@@ -0,0 +1,110 @@
...
+
+patch(ExpenseListController.prototype, {
+    displayCreateReport() {
+        const usesSampleData = this.model.useSampleModel;
+        const records = this.model.root.records;
+        return (
+            !usesSampleData &&
+            records.length &&
+            records.some((record) => record.data.state === "draft") &&
+            records.every((record) => !record.data.sheet_id)
```
carlos-lopez-tecnativa: Sorry, but I meant that if some records do not have sheet_id set, the button can be visible. However, if all records have sheet_id set, the button should not be displayed.

Currently, the button is visible only when none of the records have sheet_id set. Please check the condition again.
Image

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):
LGTM, thanks for the work and the effort to recover this feature.

victoralmau:
Ping @pedrobaeza

hr_expense_sheet/models/hr_expense_sheet.py
```diff
@@ -0,0 +1,798 @@
...
+    def _compute_untaxed_amount(self):
+        for sheet in self:
+            sheet.untaxed_amount = sum(sheet.expense_line_ids.mapped("untaxed_amount"))
+
+    @api.depends("expense_line_ids.amount_residual")
+    def _compute_amount_residual(self):
+        for sheet in self:
+            sheet.amount_residual = sum(
+                sheet.expense_line_ids.mapped("amount_residual")
+            )
```
ruuter, line 318: The amount_residual on expense is the residual of related account_move_id. So this makes the sheet residual equal account move residual x len(expenses).

hr_expense_sheet/migrations/19.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,7 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo.addons.hr_expense_sheet.hooks import pre_init_hook
+
+
+def migrate(env, version):
```
ruuter: First argument passed to migrate is cr not env.

ruuter: Note also, that currently migration still fails due to unique constraint "ir_act_window_path_unique". The sheet views already exists in 18.0 database. The XMLids should be renamed from hr_expense to hr_expense_sheet in pre-migration.

victoralmau: The `env` argument is correct; you can find similar examples in other migration scripts (although it's true that `cr` was used in previous versions).

pedrobaeza: But that argument is put when you use the decorator `@openupgrade.migrate()`, which doesn't seem the case.

carlos-lopez-tecnativa: env is correct if you decorate the method with `@openupgrade.migrate()`. However, in this case, the method is not decorated, so you should either preserve the original signature expected by Odoo or decorate it as mentioned above.

hr_expense_sheet/models/hr_expense_sheet.py
victoralmau: Added an explanation to the method

hr_expense_sheet/hooks.py
carlos-lopez-tecnativa, line 1: I hadn't tested this script before, but I have now tested it locally by simulating a database migration from V18. I noticed some inconsistencies.

I created some records in V18, backed up the database, restored it in V19, and ran OpenUpgrade. Note that this module is not installed automatically, so after the migration I installed the module manually, which triggered the `pre_init_hook`.

**In `hr.expense`**

1. I lost the `sheet_id` value.
2. The expense with a value of 200 was in the `approved` state in V18, but in V19 the record is in the `posted` state. I think this happens because OpenUpgrade doesn't know about this module, so it tries to map the states. However, this module should restore the states correctly.

See the OpenUpgrade migration script here:
https://github.com/OCA/OpenUpgrade/blob/c402d59cd518dadbabd7d6410d488ee3e81e795c/openupgrade_scripts/scripts/hr_expense/19.0.2.1/end-migration.py#L7

Image

**In `hr.expense.sheet`**

The states are not mapped correctly.

Additionally, I think some relevant information was lost when the [payment_state](https://github.com/odoo/odoo/blob/219026e958d9a6d5cbe163851f09b39c3a15b5d8/addons/hr_expense/models/hr_expense_sheet.py#L167) field was dropped. Why was it removed?
Image

NOTE: I didn't encounter any errors in the actions path as mentioned by @ruuter. Please provide the traceback.

hr_expense_sheet/hooks.py
```diff
@@ -0,0 +1,93 @@
...
+    env.cr.execute("""
+        SELECT EXISTS (
+            SELECT 1
+            FROM hr_expense
+            WHERE former_sheet_id IS NOT NULL
+        )
+    """)
+    has_former_sheet = env.cr.fetchone()[0]
+    if not has_former_sheet:
+        return
```
carlos-lopez-tecnativa, line 18: I understand that OpenUpgrade renames the `sheet_id` field to `former_sheet_id`, which is fine as a backup mechanism. However, this module should rename it back to its original name. Otherwise, a new column is created, which is why I lost the value during my test.
https://github.com/OCA/OpenUpgrade/blob/c402d59cd518dadbabd7d6410d488ee3e81e795c/openupgrade_scripts/scripts/hr_expense/19.0.2.1/pre-migration.py#L12

carlos-lopez-tecnativa (changes requested):

hr_expense_sheet/hooks.py
victoralmau: Thank you for the clarification. I have modified `pre_init_hook` to create the `sheet_id` column and set its value to the appropriate value (the one from the `former_sheet_id` column).

hr_expense_sheet/hooks.py
victoralmau: The `sheet_id` field retains the correct value (`https://github.com/OCA/hr-expense/pull/378#discussion_r4069595262`)

The `payment_state` field in `hr.expense.sheet` has been restored

A condition has been added for the sheet's status if any expenses are posted and others are paid.

carlos-lopez-tecnativa: This part is still pending. Please review it:
> **In `hr.expense.sheet`**
> 
> The states are not mapped correctly.

Image

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (changes requested):
This part is still pending. Please review it:
> **In `hr.expense.sheet`**
> 
> The states are not mapped correctly.

Image

carlos-lopez-tecnativa (approved):
I tested the migration scripts. LGTM. Please check the CI.

PR #379, opened 2026-08-13
[19.0]Update copier template to 1.44

victoralmau:
Update copier template to 1.44

Please @pedrobaeza can you review it?

@Tecnativa


pedrobaeza (approved):
The red CI is because there isn't any module yet.

